PR #1, opened 2015-01-01
Move to FoC

jadb:


jadb:
@FriendsOfCake/owners whenever you guys get a chance to review the changes made in order to adhere by the FoC requirements, let me know so I merge this branch and await for someone to add it to packagist.

ADmad:
I would prefer if the namespace was `FOC/BootstrapUI` instead of just `BootstrapUI`. Given that there will be potentially multiple bootstrap plugins it would be better to have a vendor name as top level namespace to easily distinguish this one.

jadb:
@ADmad but then wouldn't that make using this plugin different from all the rest? Crud, Search, CakePdf - all under friendsofcake vendor, none under the FOC namespace though. If this will become common to all FoC plugins, then I obviously have no objections since my argument is based on having all plugins follow some similar conventions (easier user adoption).

RE: distinguishing it among other twbs plugins, we'll do that (hopefully) by promoting it as the defacto plugin for bootstrap just like it's done already for Crud, DebugKit, AssetCompress, etc.

ADmad:
Plugins like DebugKit, Crud, CakePDF are fairly unique and it's highly unlikely someone will make plugins of similar names. The other plugins like Authenticate, Authorize, TestUtilites do use top level vendor namespace.

That said this isn't a big deal, I wouldn't really mind if vendor namespace is not use. Let's see if anyone else has an opinion on this.

josegonzalez:
I think @lorenzo thinks we shouldn't ever use top-level namespaces. I seem to recall that in an irc discussion about how to load plugins with prefixes and being told it was a bad idea.

ADmad:
I know markstory dislikes them, am not sure about @lorenzo.

lorenzo:
I do dislike them too

dereuromark:
I wish we could consistently all use the top-level namespaces, but simplify them away in our apps for all plugins by default. This would make it possible to enable them again if there ever is a conflict, but by default the annoyance of them is not present in an application.

ADmad:
I don't want this to turn into a long discussion. Considering the plugin author's preference and opinion of others I am okay with not having vendor prefix.

PR #2, opened 2015-01-06
updated bake templates

cewi:
Added templates for add, edit and view actions. Changed index template to work with the refactored bake plugin. 
The Templates are close to your old index.ctp, but now use '<%' syntax. Edit, Add and View use the sidebar for the action-links like index.ctp. (Personally I never found the amount of links useful, the original bake-templates generate. Therefore I allways delete the most of them in my own bake Templates. But for templates for general use, it's probabely better to provide the same links like the original bake templates)
BTW: What do you think about 'augmenting' the Html-Helper to add (glyph)icons in the markup for Links (and the formHelper for postLinks, too)? I made a plugin for my own purpose which adds this functionality to your plugin. See cewi/cakephp_tb_starter. If you like the idea, the helpers could be included in your plugin.

Thanks for your work!


src/Template/Bake/Template/index.ctp
```diff
@@ -0,0 +1,94 @@
...
+?>
+<ul class="nav nav-sidebar">
+    <li><?= $this->Html->link(__('New <%= $singularHumanName %>'), ['action' => 'add']); ?></li>
+    <%
+    $done = [];
+    foreach ($associations as $type => $data):
+        foreach ($data as $alias => $details):
+            if ($details['controller'] != $this->name && !in_array($details['controller'], $done)):
+                %>
+    <li><?= $this->Html->link(__('List <%= Inflector::humanize($details["controller") %>'), ['controller' => '<%= $details["controller"] %>', 'action' => 'index']); ?></li>
```
DSchalla: You are missing an ']' after the "controller", this way the bake will fail. Otherwise the bake worked fine and seems to be complete.

jadb:
@cewi thanks for the contribution! I should be merging it real soon.

I definitely have plans to augment this plugin with all of bootstrap's components and CSS stuff (just not enough time). I will look at your plugin and unless it should be done differently (widgets vs stringtemplates), I would gladly add it!

@DSchalla thanks for the rapid fixes bake guy ;)

PR #3, opened 2015-01-09
Add control-label class to formgroups to automagically support validatio...

bravo-kernel:
Adds the BS3 "control-label" class to all generated formgroup labels which makes the label respect the validation state of the formgroup (http://getbootstrap.com/css/#forms-control-validation). Should be non-obtrusive.

Now needs:

    echo $this->Form->input('url', [
        'label' => ['class' => 'control-label']
    ]); 

After PR:

    echo $this->Form->input('url');

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -84,7 +84,7 @@ class FormHelper extends Helper
             'prepend' => null,
             'append' => null,
             'type' => null,
-            'label' => null,
+            'label' => ['class' => 'control-label'],
             'error' => null,
             'required' => null,
             'options' => null,
--- a/tests/TestCase/View/Helper/FormHelperTest.php
+++ b/tests/TestCase/View/Helper/FormHelperTest.php
@@ -67,7 +67,7 @@ class FormHelperTest extends TestCase
         $result = $this->Form->input('title');
         $expected = [
             'div' => ['class' => 'form-group'],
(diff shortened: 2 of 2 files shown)


ADmad:
You need to update the tests too.

bravo-kernel:
All tests now pass except for the assert on [line 404](https://github.com/FriendsOfCake/bootstrap-ui/blob/05aa194b1089c38b5a23fdd9ffcab086b279adc6/tests/TestCase/View/Helper/FormHelperTest.php#L404). Some assistance would be appreciated. 

The problem seems to be caused by the code somehow producing **two separate class attributes** for the label:

    <div class="form-group">
        <label class="col-md-2" class="control-label" for="title">Title</label>
        <div class="col-md-10">
             <input type="text" name="title" required="required" id="title" class="form-control">
        </div>
    </div>

jadb:
@bravo-kernel thanks man for the effort. I have looked into the changes and read back the documentation for Bootstrap. I noticed two major issues with your implementation that forced me to re-write my own and push it:

- Labels for inputs of type `checkbox` or `radio` do not use the `control-label` class
- Horizontal forms (in your failing test) need 2 classes by default (`col-md-:n` and `control-label`)

In any case, the bug you have raised is now fixed. Thanks for not just pointing it out but also trying to fix it :+1:

bravo-kernel:
Thanks for solving this the proper way @jadb, good job :+1:

PR #5, opened 2015-01-12
Adds has-feedback class to form-groups

bravo-kernel:
Adds the "has-feedback" class to form-group elements to support optional feedback icons out-of-the-box like this example from the [BS3 page on form validation](http://getbootstrap.com/css/#forms-control-validation):

    <div class="form-group has-success has-feedback">
      <label class="control-label" for="inputSuccess2">Input with success</label>
      <input type="text" class="form-control" id="inputSuccess2" aria-describedby="inputSuccess2Status">
      <span class="glyphicon glyphicon-ok form-control-feedback" aria-hidden="true"></span>
      <span id="inputSuccess2Status" class="sr-only">(success)</span>
    </div>

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -21,8 +21,8 @@ class FormHelper extends Helper
         $this->_defaultConfig['errorClass'] = null;
         $this->_defaultConfig['templates'] = array_merge($this->_defaultConfig['templates'], [
             'error' => '<div class="text-danger">{{content}}</div>',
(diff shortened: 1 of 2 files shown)


bravo-kernel:
Oops, this one was a bit too enthusiastic. 

> Feedback icons only work with textual <input class="form-control"> elements.

The PR now sets the feedback for checkboxes too which is not desired.

jadb:
Ya - sorry I will have to close this PR as it forces the `has-feedback` class onto all created inputs - really not the desired effect.

Please open an issue instead and I will look into it.

PR #11, opened 2015-02-21
Remove flash component

AD7six:
It's not actually necessary if a theme is used - see https://github.com/cakephp/cakephp/pull/5936.

Alternatively the elements could/should be copied to the app (instead of using a theme).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Transparently use [Twitter Bootstrap 3][twbs3] with [CakePHP 3][cakephp].
 
 ## What's included?
 
-- FlashComponent + view elements (types: error/danger, info, success, warning)
+- FlashComponent + elements (types: error/danger, info, success, warning)
 - FormHelper (types: vertical, inline, horizontal | inputs: text, textarea, select, checkbox, radio)
 - Widgets (button, textarea)
 - Sample layouts (cover, signin, dashboard)
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Transparently use [Twitter Bootstrap 3][twbs3] with [CakePHP 3][cakephp].
 
 ## What's included?
 
-- FlashComponent + elements (types: error/danger, info, success, warning)
+- FlashComponent + elements (types: error, info, success, warning)
 - FormHelper (types: vertical, inline, horizontal | inputs: text, textarea, select, checkbox, radio)
 - Widgets (button, textarea)
 - Sample layouts (cover, signin, dashboard)
@@ -27,11 +27,6 @@ Using [Composer][composer]:
 ```
 composer require friendsofcake/bootstrap-ui:dev-master
 ```
-
-Because this plugin has the type `cakephp-plugin` set in its own `composer.json`,
-[Composer][composer] will install it inside your /plugins directory, rather than
-in your `vendor-dir`. It is recommended that you add /plugins to your `.gitignore`
-file and here's [why][composer:ignore].
 
 You then need to load the plugin. In `boostrap.php`, something like:
 
--- a/src/Controller/Component/FlashComponent.php
+++ b/src/Controller/Component/FlashComponent.php
@@ -18,10 +18,6 @@ class FlashComponent extends Component
      */
     public function __call($name, $args)
     {
-        if ($name === 'danger') {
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -42,7 +42,7 @@ You then need to load the plugin. In `boostrap.php`, something like:
 For a complete setup, add the following to your `App\Controller\AppController`:
 
 ```php
-public $components = ['BootstrapUI.Flash'];
+public $theme = 'BootstrapUI';
```
ADmad: I would prefer if we suggested all view related setting be done through AppView.

AD7six: It's probably better to bake them into the app, rather than add the overhead of a theme for all views, that are going to be used for flash elements only - but there aren't any bake templates yet (I do that).

README.md
```diff
@@ -13,7 +13,7 @@ Transparently use [Twitter Bootstrap 3][twbs3] with [CakePHP 3][cakephp].
 
 ## What's included?
 
-- FlashComponent + view elements (types: error/danger, info, success, warning)
+- View elements (types: error (danger), info, success, warning)
```
ADmad: "Flash elements" might be better?

README.md
ADmad: Baking might be better but if we want to stay lazy we could replace the FlashComponent with FlashHelper with code similar to what i had added in core PR.

jadb: Doesn't this change mean that one can't use any other theme for their views now? Does it also imply that one needs to create their views in a different folder (since they are 'themed')? 

I haven't worked with themes on cake3 but just wondering as this plugin, despite only being related to the view layer, isn't a theme per se, more a set of helpers to get you going with Bootstrap.

ADmad: > Doesn't this change mean that one can't use any other theme for their views now?

Why would one need to use other theme? I always viewed the plugin itself a glorified theme.

> Does it also imply that one needs to create their views in a different folder (since they are 'themed')?

No.

jadb: It definitely could be a theme type of plugin but then we really need to have more pre-themed stuff ready. 

As I said, thus far it only has some helpers; but it doesn't for example have bake templates for grids, views, forms, etc. By taking the 'theme' spot in an app, I believe the plugin then **has** to offer those at least, as in *not optional*, or it would suck as a 'theme'. What do you think?

AD7six: To be honest, I find this plugin confusing.

It's sort of a theme, but you don't use it as a theme. I've reverted removing the flash component, addressing one of the _other_ things I found confusing (the use of symlinks, unconventional flash element names, and forcing the use of elements in the plugin that don't exist (if a user were to do `$this->Flash->foo('message')).

AD7six:
I think in general we should have a clearer goal for this plugin.

I would suggest to go more in the direction of: You bake everything you need.

src/Controller/Component/FlashComponent.php
```diff
@@ -18,11 +18,15 @@ class FlashComponent extends Component
      */
     public function __call($name, $args)
     {
-        if ('error' == $name) {
-            $name = 'danger';
+        if ($name === 'danger') {
```
ADmad: Don't see any good reason why anyone would call Flash->danger() instead of Flash->error(). Just because bootstrap chose a silly name for the css class doesn't me one has to use it in php code too. In general no other name besides besides the standard error levels makes sense to me.

Once this is merged i'll replace this component with helper. The element name jugglery done below belongs in View layer imo.

AD7six: Only to maintain BC if someone is already doing that.

ADmad: Until we make a release we don't really have to maintain BC :stuck_out_tongue:

AD7six: Fair =). I removed the snowflake.

jadb: :+1:

jadb:
@AD7six I totally agree and I like the bake idea. Since I never had worked with bake, I just didn't have any added. 

I am not against this being a 'theme' type of plugin and people can extend on it for even more customizations, but as we seem to all agree on, it needs some love (mainly bake stuff) before it can be defined as such.

PR #12, opened 2015-02-22
Remove flash component

ADmad:
- Added FlashHelper
- Added Flash/default.ctp  (with class handling similar to what's done in the app's element)
- Removed FlashComponent as it's responsibilities are now taken over by the helper

Scope for improvement:

- @AD7six suggested making the alerts dismissable, which is a matter of updating the attributes of the alert divs.


README.md
```diff
@@ -37,8 +37,10 @@ You then need to load the plugin. In `boostrap.php`, something like:
 For a complete setup, add the following to your `App\Controller\AppController`:
 
 ```php
-public $components = ['BootstrapUI.Flash'];
-public $helpers = ['BootstrapUI.Form'];
+public $helpers = [
+    'Form' => ['className' => BootstrapUI.Form'],
+    'Flash' => ['className' => 'BootstrapUI.Flash']
+];
```
jadb: Since everything is on the view layer now (getting closer to a theme type of plugin, yay!), would you prefer suggesting the helpers to be loaded in `AppView` instead?

src/View/Helper/FlashHelper.php
```diff
@@ -0,0 +1,62 @@
...
+            ));
+        }
+        $flash = $options + $flash;
+        $this->request->session()->delete("Flash.$key");
+
+        $element = $flash['element'];
+        if (strpos($element, '.') === false &&
+            in_array($element, ['Flash/default', 'Flash/error', 'Flash/info', 'Flash/success', 'Flash/warning'])
+        ) {
+            $element = 'BootstrapUI.' . $element;
```
AD7six: shouldn't the element always be "default" now since it resolves css-classes itself?

src/Template/Element/Flash/default.ctp
```diff
@@ -0,0 +1,11 @@
+<?php
+$class = 'alert';
+if (!empty($params['class'])) {
```
AD7six: picky point, would be "nicer" I think if class, if it's present, is expected to be an array.

i.e.:

    if (!empty($params['class'])) {
        $classes = (array)$params['class'];
    } else {
        $classes[] = 'alert-info';
    }
    $classes[] = 'alert';
    $classes = array_unique($classes);
    ?>
    <div class="<?= h(implode(' ', $classes)); ?>">
    
Makes conditional logic slightly easier, such as:

    <div class="<?= h(implode(' ', $classes)); ?>">
    <?php if (in_array('alert-dismissable', $classes)): ?>
        <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
    <?php endif; ?>

ADmad: Using array of classes does make more sense. 

Regarding making an alert dismissable how about we add an option `dismissable` (default true) to the FlashHelper?  If true the helper can automatically add `alert-dismissable` to the classes list. This will help keep controller code DRY.

AD7six: I thought the same but didn't want to drown you in feature-requests =).

README.md
ADmad: Yup, I'll update the docs later.

ADmad:
Updates done as discussed. The helper code is a bit uglyish but does the job.

src/View/Helper/FlashHelper.php
```diff
@@ -0,0 +1,85 @@
...
+        $this->request->session()->delete("Flash.$key");
+
+        $element = $flash['element'];
+        if (strpos($element, '.') === false &&
+            preg_match('#Flash/(default|success|error|info|warning)$#', $element, $matches)
+        ) {
+            $class = $matches[1];
+            if ($class == 'default') {
+                $class = 'info';
+            }
```
jadb: :+1:

src/View/Helper/FlashHelper.php
```diff
@@ -0,0 +1,85 @@
...
+
+        return $this->_View->element($element, $flash);
+    }
+
+    /**
+     * Event listeners.
+     *
+     * @return array
+     */
+    public function implementedEvents()
```
jadb: Extend `Cake\View\Helper\FlashHelper` and get rid of the extra lines?

src/Template/Element/Flash/default.ctp
```diff
@@ -0,0 +1,13 @@
+<?php
+$class = array_unique((array)$params['class']);
+$message = h($message);
+
+if (in_array('alert-dismissible', $class)) {
+    $button = <<<BUTTON
+<button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+BUTTON;
```
jadb: Any reason for using raw HTML markup instead of a tag generation helper method (i.e. `Form::button()`)?

README.md
```diff
@@ -34,12 +34,16 @@ You then need to load the plugin. In `boostrap.php`, something like:
 \Cake\Core\Plugin::load('BootstrapUI');
 ```
 
-For a complete setup, add the following to your `App\Controller\AppController`:
+For a complete setup, add the following to your `App\View\AppView`:
 
 ```php
-public $components = ['BootstrapUI.Flash'];
-public $helpers = ['BootstrapUI.Form'];
 public $layout = 'BootstrapUI.default';
```
jadb: Can one change the layout value in a controller if only defined in `AppView`? I usually define `$layout` in `AppController` and the load the helpers with `AppView`. Maybe am doing it wrong?

src/View/Helper/FlashHelper.php
ADmad: I would rather duplicate two lines and avoid an extra filesystem hit :)

README.md
ADmad: Layout value you set in controller will still override the one in AppView.

src/Template/Element/Flash/default.ctp
ADmad: Unless there are variables involved straight up HTML is way more convenient and readable than nested calls to helper methods :)

README.md
jadb: Nice, good to know!

src/Template/Element/Flash/default.ctp
jadb: Fair enough.

src/View/Helper/FlashHelper.php
jadb: :100:

PR #14, opened 2015-02-23
Correct issue with double inputs on checkbox and radio

dakota:
Checkbox and radio inputs are included into the `{{label}}` element, so having a `{{input}}` causes double rendering of the checkbox/radio.


dakota:
Issue is deeper than I thought, and relates to multicheckboxes. Will open another PR.

PR #15, opened 2015-02-23
Correct issue with double inputs on multicheckbox

dakota:
Checkboxes are already rendered inside the `{{label}}` element.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -119,6 +119,9 @@ class FormHelper extends Helper
             case 'select':
             case 'multiselect':
             case 'textarea':
+                if (isset($options['multiple']) && $options['multiple'] === 'checkbox') {
+                    $this->templates(['checkboxWrapper' => '<div class="checkbox">{{label}}</div>']);
+                }
                 break;
 
             default:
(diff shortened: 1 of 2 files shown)


ADmad:
Tests needed.

dakota:
Tests are for sissies, just wing it :P

Will push them just now.

dakota:
Added a test.

src/View/Helper/FormHelper.php
```diff
@@ -120,6 +120,10 @@ public function input($fieldName, array $options = [])
             case 'select':
             case 'multiselect':
             case 'textarea':
+                if (isset($options['multiple']) && $options['multiple'] === 'checkbox') {
```
jadb: Maybe this should be limited to the `select` case? By default I think cake will assign the `select` to `$options['type']` given there is an array for `$options['options']`.

dakota: You are correct. I've updated the PR.

ADmad:
@dakota You need to rebase, the PR is no longer mergeable.

dakota:
@ADmad Rebased

jadb:
Thanks @dakota :)

PR #17, opened 2015-02-24
Fix styles, Bootstrap 3 uses danger instead of error

sarrala:
ref: http://getbootstrap.com/migration/


jadb:
Thanks @sarrala - it was removed not long ago and I couldn't find the reason behind it, so I'll wait on @AD7six to approve this one.

For reference: https://github.com/FriendsOfCake/bootstrap-ui/commit/c4aa55c7cc8748e8dd8ece46b9b01e190e64b6c9

ADmad:
Css class names should not dictate method names / error types. I had missed out ensuring `alert-danger` to message of type error, #18 takes care of that.

ADmad:
#18 has been merged.

Issue #9, opened 2015-02-03
Checkbox for horizontal form generating wrong HTML


aceat64:
When using `$this->Form->input('admin')` I get the following HTML output:
```
<div class="form-group">
    <div class="col-md-offset-2 col-md-10">
        <input type="hidden" name="admin" value="0">
        <label for="admin">
            <input type="checkbox" name="admin" value="1" id="admin" class="form-control">
            Admin
        </label>
    </div>
</div>
```

When I should be getting:
```
<div class="form-group">
    <div class="col-md-offset-2 col-md-10">
        <div class="checkbox">
            <label>
                <input type="hidden" name="admin" value="0"><input type="checkbox" name="admin" value="1"> Admin
            </label>
        </div>
    </div>
</div>
```

dakota:
PR now open (#21)

PR #21, opened 2015-03-09
Fix checkbox rendering

dakota:
This fixes the checkbox rendering as reporting in #9.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -67,7 +67,7 @@ class FormHelper extends Helper
             $options['templates'] += [
                 'label' => '<label class="' . $options['horizontal']['left'] . '"{{attrs}}>{{text}}</label>',
                 'formGroup' => '{{label}}<div class="' . $options['horizontal']['right'] . '">{{input}}</div>',
-                'checkboxFormGroup' => '<div class="' . $options['horizontal']['combined'] . '">{{label}}</div>',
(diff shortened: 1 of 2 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -67,7 +67,8 @@ public function create($model = null, array $options = [])
             $options['templates'] += [
                 'label' => '<label class="' . $options['horizontal']['left'] . '"{{attrs}}>{{text}}</label>',
                 'formGroup' => '{{label}}<div class="' . $options['horizontal']['right'] . '">{{input}}</div>',
-                'checkboxFormGroup' => '<div class="' . $options['horizontal']['combined'] . '">{{label}}</div>',
+                'checkboxFormGroup' => '<div class="' . $options['horizontal']['combined'] . '">
+                <div class="checkbox">{{label}}</div></div>',
```
jadb: Maybe better to use string concatenation when splitting the string?

src/View/Helper/FormHelper.php
```diff
@@ -162,8 +163,12 @@ public function input($fieldName, array $options = [])
                 $this->templates(compact('input'));
         }
 
+        if ($options['type'] !== 'checkbox' && $options['type'] !== 'radio') {
```
jadb: How about:

```php
if (!in_array($options['type'], ['checkbox', 'radio]))
```

jadb:
Looks good, thanks! :+1:

dakota:
Updated based on your comments/suggestions @jadb

PR #23, opened 2015-03-28
Fix default checkbox input

jadb:
@Schalla pointed out that the tests for the default checkbox
input were asserting a wrong markup according to [GetBootstrap].

Since I am not sure about all the other ways checkboxes can
be displayed, I left the rest (inline ones and ones used in
horinzontal forms untouched). Will revisit if they are reported
broken.

[GetBootstrap]:http://getbootstrap.com/css/#forms

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -114,6 +114,13 @@ class FormHelper extends Helper
                 if (!isset($options['inline'])) {
                     $options['inline'] = $this->checkClasses('checkbox-inline', (array)$options['label'])
                         || $this->checkClasses('radio-inline', (array)$options['label']);
+
+                    if (!$this->_align) {
+                        $this->templates([
+                            'checkboxContainer' => '<div class="checkbox">{{content}}</div>',
+                            'checkboxContainerError' => '<div class="checkbox has-error">{{content}}</div>',
+                        ]);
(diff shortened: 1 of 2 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -106,6 +114,13 @@ public function input($fieldName, array $options = [])
                 if (!isset($options['inline'])) {
                     $options['inline'] = $this->checkClasses('checkbox-inline', (array)$options['label'])
                         || $this->checkClasses('radio-inline', (array)$options['label']);
+
+                    if (!$this->_align) {
+                        $this->templates([
+                            'checkboxContainer' => '<div class="checkbox">{{content}}</div>',
```
ADmad: All this template juggling is getting ugly and will be difficult to maintain in future. We should have two config file will all templates and just select the one required whether user wants horizontal or not.

jadb: 100% - am already thinking of a way to refactor this in a cleaner way.

jadb:
Merging all recent PRs and then tagging a new patch release for 0.2 before I start work on a 0.3 (dev) to try using widgets for better organizing everything (and hopefully add a couple new features along the way). Might break the API but we're not on 1.0 yet - so any comments welcomed once I create the branch.

PR #24, opened 2015-03-28
Add {{required}} placeholder to input containers.

ADmad:
While BS does not have a "required" class for form groups, I would like to use it to add extra styling to indicate required fields. This change is unlikely to affect existing users.


jadb:
I don't mind it.

PR #26, opened 2015-03-28
Fix Error Handling to follow Bootstrap Guidelines

DSchalla:
See:
http://getbootstrap.com/css/#forms-example

When the class has-error is set all elements will use anyway red coloring, the current usage of text-danger is wrong.


jadb:
Closing this in favour of previously merged #27.

PR #27, opened 2015-03-28
Fixed Error Handling to use intended Bootstrap Elements

DSchalla:
The help-block should also be used for error-handling, since all text elements be colored red anyway and the help-block is intended to be placed under a text input unlike text-warning.


DSchalla:
For reference: http://getbootstrap.com/css/#forms-example

PR #28, opened 2015-03-28
Better default layout for horizontal.

dereuromark:
For a default cake app there is a lot of space wasted through way too big forms.
And the brake point should be "sm" not "md" IMO.

I also think horizontal should probably be the default.
But with "options" coming to globally activate this, there is probably no reason to change this.


jadb:
@dereuromark do you think I should leave this as is in master for 0.2 since they are modifiable in 0.3 and would be considered a BC if merged into current release?

ADmad:
I would leave it as is since customizing it will be possible in 0.3

dereuromark:
I agree with @ADmad

jadb:
Ok, thanks for confirming guys! Closing then.

PR #30, opened 2015-03-28
Minor merge improvement.

dereuromark:
Should all be string based keys.


dereuromark:
@jadb Should be good to merge.

Issue #29, opened 2015-03-28
type radio broken


dereuromark:
```php
echo $this->Form->input('gender', ['type' => 'radio', 'options' => ['Male', 'Female']]);
```

![unbenannt](https://cloud.githubusercontent.com/assets/39854/6882362/399a56ce-d582-11e4-981b-d491ef319965.PNG)

The radio button appears twice per value

jadb:
I think this is resolved in the current 0.3 (dev). Keeping it open for now as it might need fixing on 0.2 if there are BC in 0.3 (for people who will stick to 0.2 for now).

Issue #20, opened 2015-03-03
twitter/bootstrap package integration


danielzzz:
hi, it would be great if this package used twitter/bootstrap from composer and connected to the styles provided by this package (if possible). thanks. dan

danielzzz:
it might be useful to have the same with jquery and any other dependency, too

dakota:
In my opinion, things like bootstrap, jquery, etc. are better served by using a tool designed for web assets such as bower.

AD7six:
Using composer to install [https://github.com/twbs/bootstrap](bootstrap) might make sense - I'd rather adding this plugin didn't force-include it though as it's making a choice for you.

> it might be useful to have the same with jquery and any other dependency, too

I have never heard of installing jquery (or any other js/css/not-php lib) via compser... do people really do that?

ADmad:
Even if we did include bootstrap as dependency (not that i am in favor of that) it would get installed in vendor directory while cake needs the assets somewhere under webroot. Using bower as @dakota suggested would be a better option.

danielzzz:
I said composer, but yeah, bower would be a better option. 
but the point is, it would be great to have it working out of the package without having to manually add extra files.
Maybe using CDN urls for js and css dependencies would solve this problem.

jadb:
We could add some copy/paste instructions to the README file but the plugin can't (and shouldn't) install bower packages for the end-user.

ADmad:
@jadb Copy / paste instructions sounds good.

Issue #7, opened 2015-01-13
Add form (validation) feedback


bravo-kernel:
As requested. Please consider implementing these additions so generated forms will support validation feedback out-of-the-box.
# Feedback icons

To add <em>optional feedback icons</em> support the "has-feedback" class has to be added to (only the) textual input elements with class "form-control".
### Vertical

HTML example from the Bootstrap 3 page:

```
<div class="form-group has-success has-feedback">
  <label class="control-label" for="inputSuccess2">Input with success</label>
  <input type="text" class="form-control" id="inputSuccess2" aria-describedby="inputSuccess2Status">
  <span class="glyphicon glyphicon-ok form-control-feedback" aria-hidden="true"></span>
  <span id="inputSuccess2Status" class="sr-only">(success)</span>
</div>
```

Working CakePHP generated example:

```
<div class="form-group has-error has-feedback">
    <label class="control-label" for="url">Url</label>
    <input name="url" required="required" id="url" class="form-control" type="text">
    <span class="fa fa-times form-control-feedback" aria-hidden="true"></span>
</div>
```

Produces something like this:

![bsui-feedback-vertical](https://cloud.githubusercontent.com/assets/230500/5726858/0020347a-9b5f-11e4-954d-6725d94a65f4.png)

> please note: that checkbox should not have the "has-feedback" class.
### Horizontal

The horizontal form needs a slightly different layout, as shown in the Bootstrap code below:

```
  <div class="form-group has-success has-feedback">
    <label class="control-label col-sm-3" for="inputSuccess3">Input with success</label>
    <div class="col-sm-9">
      <input type="text" class="form-control" id="inputSuccess3" aria-describedby="inputSuccess3Status">
      <span class="glyphicon glyphicon-ok form-control-feedback" aria-hidden="true"></span>
      <span id="inputSuccess3Status" class="sr-only">(success)</span>
    </div>
  </div>
```

Currently generated CakePHP code:

```
<div class="form-group has-feedback has-error">
    <label class="control-label col-md-2" for="url">Url</label>
    <div class="col-md-10">
        <input name="url" required="required" id="url" class="form-control" type="text">
    </div>
    <span class="fa fa-times form-control-feedback" aria-hidden="true"></span>
</div>
```

Now produces something like: 

![bsui-feedback-horizontal](https://cloud.githubusercontent.com/assets/230500/5726611/10197b9a-9b5d-11e4-89c3-a35e3feee99f.png)
# Help block

[...]

PR #33, opened 2015-03-29
0.3

jadb:
I'll fill this later on with the goals of this refactor.

Tasks:
- [ ] ~~#7~~ *moved to future milestone*
- [x] #20
- [x] #29
- [ ] ~~a style guide ala [github primer](http://primercss.io/scaffolding/) for most of the existing features - this would serve as both a documentation for the developer (client, designer, etc.) and a visual test for us on the long term.~~ *moved to future milestone*
- [x] update composer command in README to 0.3
- [x] Fix placement of help block for horizontally aligned forms (similar to what was done for error message).


src/View/Helper/FormHelper.php
```diff
@@ -276,4 +232,124 @@ protected function _parseOptions($fieldName, $options)
         }
         return $options;
     }
+
+    /**
+     * Form alignement detector/switcher.
+     *
+     * @param  array $options Options.
```
ADmad: 2 spaces after `@param` instead of 1.

src/View/Helper/FormHelper.php
```diff
@@ -276,4 +232,124 @@ protected function _parseOptions($fieldName, $options)
...
+
+        if (is_array($options['align'])) {
+            $this->_grid = $options['align'];
+            $options['align'] = 'horizontal';
+        } elseif ($options['align'] === 'horizontal') {
+            $this->_grid = $this->config('grid');
+        }
+
+        if (!in_array($options['align'], ['default', 'horizontal', 'inline'])) {
+            throw new InvalidArgumentException();
```
ADmad: Add an exception message please.

ADmad:
Adding a formhelper method/widget for "static" type might also be useful http://getbootstrap.com/css/#forms-controls-static  Usage would be `Form->input('field', ['type' => 'static'])`. We could provide the option whether or not to include hiddenfield for it.

jadb:
@ADmad I really like that suggestion. Will look into it.

README.md
```diff
@@ -14,11 +14,12 @@ Transparently use [Twitter Bootstrap 3][twbs3] with [CakePHP 3][cakephp].
 ## What's included?
 
 - FlashComponent + elements (types: error, info, success, warning)
-- FormHelper (types: vertical, inline, horizontal | inputs: text, textarea, select, checkbox, radio)
-- Widgets (button, textarea)
+- FormHelper (types: default, inline, horizontal)
```
ADmad: Should be "align" instead of "types". `type` is already an option of create() so can cause confusion.

src/View/Helper/PaginatorHelper.php
```diff
@@ -0,0 +1,96 @@
...
+            $options['class'] .= " {$class}-{$options['size']}";
+        }
+
+        $options += [
+            'before' => '<ul class="' . $options['class'] . '">',
+        ];
+
+        unset($options['class'], $options['size']);
+
+        if (isset($options['prev'])) {
```
ADmad: How about including prev and next links by default?  Or at least allow using ['prev' => true] in options which would generate the prev link using `&laquo;` as shown here http://getbootstrap.com/components/#pagination

src/View/Helper/PaginatorHelper.php
```diff
@@ -0,0 +1,96 @@
...
+    public function __construct(View $View, array $config = [])
+    {
+        $this->_defaultConfig['templates'] = [
+            'nextActive' => '<li class="next"><a rel="next" aria-label="Next" href="{{url}}">' .
+                            '<span aria-hidden="true">{{text}}</span></a></li>',
+            'nextDisabled' => '<li class="next disabled"><a href=""><span aria-hidden="true">{{text}}</span></a></li>',
+            'prevActive' => '<li class="prev"><a rel="prev" aria-label="Previous" href="{{url}}">' .
+                            '<span aria-hidden="true">{{text}}</span></a></li>',
+            'prevDisabled' => '<li class="prev disabled"><a href=""><span aria-hidden="true">{{text}}</span></a></li>',
+            'current' => '<li class="active">{{text}} <span class="sr-only">(current)</span></li>',
```
ADmad: This actually needs to be `<li class="active"><span>{{text}} <span class="sr-only">(current)</span></span></li>`  http://getbootstrap.com/components/#disabled-and-active-states

[insert obligatory comment about BS and html tag soup]

src/View/Helper/FormHelper.php
```diff
@@ -276,4 +234,124 @@ protected function _parseOptions($fieldName, $options)
...
+            throw new InvalidArgumentException('Invalid `align` option value.');
+        }
+
+         $this->_align = $options['align'];
+
+         unset($options['align']);
+
+        if ($this->_align === 'default') {
+            $options['templates'] += [
+               'checkboxContainer' => '<div class="checkbox">{{content}}{{help}}</div>',
```
ADmad: Overriding these templates in the method means user cannot customize the if required. All alternate templates should also be in the $_template so that once it's merged with config they can be easily overridden during forrmhelper initialization or method call.

We can have an extra level of nesting in $_templates if required to save each set per alignment. E.g.

```
$_templates = [
    'default' => ['checkboxContainer' => '...', 'checkboxContainerError' => '...'],
    'inline' => ['label' => '...']
]
```

jadb: Not overriding here - appending `+=`. Am I missing something?

src/View/Helper/FormHelper.php
```diff
@@ -25,64 +52,50 @@ class FormHelper extends Helper
...
-        ]);
+        $this->_defaultConfig = [
+            'align' => 'default',
+            'errorClass' => null,
+            'grid' => [
+                'left' => 2,
+                'middle' => 6,
+                'right' => 4
+            ],
+            'templates' => array_merge($this->_defaultConfig['templates'], $this->_templates),
```
dereuromark: the merging can also use´+` operator here as the keys must be stringish

src/View/Helper/FormHelper.php
ADmad: Sorry I misread the code. 

Still perhaps for easier management and better readability it might be better if these extra template sets were keep in a var instead of having them inline in the method.

ADmad:
@jadb This looks good. You have anything else on the list to be still added (besides the "static" input type i proposed) ?

composer.json
```diff
@@ -29,6 +29,7 @@
     }
     , "require-dev": {
         "phpunit/phpunit": "4.1.*"
+        , "cakephp/cakephp-codesniffer": "2.*"
```
ADmad: This is a big eye sore for me. Can you please switch to having comma at end of statement for all foc projects :smile:

jadb: I will - was gonna make my life easier and do so anyways. Just because it's JSON produced by PHP but generally speaking, [comma should be first](https://docs.npmjs.com/misc/coding-style#comma-first).

dereuromark: I can understand the reasoning behind it, but it's as useful to the human mind as yoda conditions :P

jadb:
Ok, I like the progress thus far. I do have a couple things I'd still like to find time to finish:

- [ ] #7 
- [ ] #20
- [ ] a style guide ala [github primer](http://primercss.io/scaffolding/) for most of the existing features - this would serve as both a documentation for the developer (client, designer, etc.) and a visual test for us on the long term.
- [ ] update composer command in README to 0.3

src/View/Helper/FormHelper.php
```diff
@@ -114,29 +128,15 @@ public function input($fieldName, array $options = [])
...
+                    $options['label'] = $this->injectClasses($options['type'] . '-inline', (array)$options['label']);
                 }
 
-                unset($options['inline']);
-
                 break;
-
             case 'select':
                 if (isset($options['multiple']) && $options['multiple'] === 'checkbox') {
                     $this->templates(['checkboxWrapper' => '<div class="checkbox">{{label}}</div>']);
```
jadb: I'd like to get rid of this also. The template reset in this method is only required because of this change that hasn't been moved yet.

README.md
```diff
@@ -114,7 +133,18 @@ will render this HTML:
 ### Horizontal Form
 
 ```php
-echo $this->Form->create($article, ['horizontal' => true]);
+echo $this->Form->create($article, ['align' => [
+    'sm' => [
+        'left' => 6
+        'middle' => 6
+        'right' => 12,
```
ADmad: This adds upto 24 cols, probably incorrect.

ADmad:
@jadb All tasks have been care of except 2. 

I would suggest pushing #7 to 0.4, since the feature will rely on adding js and getting the placement of icon correctly for various input types will be tricky. 

A visual guide is definitely something we should help but imo that can be done after 0.3.0 release.

src/View/Helper/FormHelper.php
```diff
@@ -276,4 +302,122 @@ protected function _parseOptions($fieldName, $options)
...
+        if ($this->_align === 'inline') {
+            $options['templates'] += [
+               'label' => '<label class="sr-only"{{attrs}}>{{text}}</label>',
+               'inputContainer' => '{{content}}'
+            ];
+            return $options;
+        }
+
+        $offsetedGridClass = implode(' ', [$this->_gridClass('left', true), $this->_gridClass('middle')]);
+        $options['templates'] += [
```
ADmad: I am still not happy about this. While this allows you to customize the templates through create() or input() calls, one can't do so using config when loading helper. If i do `$this->loadHelper('Form', ['className' => 'BootstrapUI.Form', 'templates' => [...]]` my templates will get overriden by the ones set by this code.

Solution to this would be what i had proposed earlier, having a set of templates per alignment type which can be modified through config and then swapped here based on form alignment type.

ADmad: Here's an possible implementation for what i described above https://gist.github.com/ADmad/e6891baabae6bec030b5

jadb: :100:  That's exactly what I had in mind when I started splitting things up!

Am not sure about the key name `templatesSet`, maybe just appending them to `templates` by alignment (i.e. `templates.horizontal`, `templates.vertical`) and the default, would use the default `templates`. Or is there a reason for the separate key that I have overlooked?

jadb:
@ADmad thanks!!

I am ok with moving the remaining tasks to 0.4 and get this merged. It's big enough as it is and would love to get feedback from users.

Let's get the last proposed changes in and merge this PR!

src/View/Helper/FormHelper.php
ADmad: Can't have it under $_config['templates'] as the StringTemplate class "compiles" the templates and barfs when if finds nested array.

jadb: Ah, ok. Didn't know that. Let's apply this patch then!

PR #38, opened 2015-04-07
Fix error when using custom templates file.

### LINKED ISSUE #37: Loading of a string template is broken in 0.3.0
Loading a template config file (e.g. `config/search_templates.php`) is currently broken in 0.3.0:

```php
echo $this->Form->create(null, ['type' => 'get', 'class' => 'search', 'templates' => 'search_templates']);
```

results in

```
Error: Unsupported operand types 
File /var/www/html/vendor/friendsofcake/bootstrap-ui/src/View/Helper/FormHelper.php 
Line: 367
```

Temporary work-around is to load the template file after create with

```php
$this->Form->templates('search_templates');
```

This may be the best course of action to fix the problem.

ADmad:
Closes #37 

I chose to directly load the templates file instead of using templater function because that would require backing up / restoring templates, plus StringTemplate::load() also compiles templates which would be unnecessary overhead as the compilation is going to be done later in core's FormHelper::create()  anyway.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -2,6 +2,7 @@
 
 namespace BootstrapUI\View\Helper;
 
+use Cake\Core\Configure\Engine\PhpConfig;
 use Cake\View\Helper\FormHelper as Helper;
 use Cake\View\View;
 use InvalidArgumentException;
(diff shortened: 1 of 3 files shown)


dakota:
That does the trick :+1: 

Good idea on loading it directly, that was what I was stuck with (How to load it without double compiling or overwriting templates)

PR #44, opened 2015-04-13
fixes #43

### LINKED ISSUE #43: Error
I have an Error when i pass to cake 3.1 

 Error: Unsupported operand types 
File /var/www/timesheets/vendor/friendsofcake/bootstrap-ui/src/View/Helper/FlashHelper.php 
Line: 55

makamo:
change += to array_merge()

### DIFF
--- a/src/View/Helper/FlashHelper.php
+++ b/src/View/Helper/FlashHelper.php
@@ -52,7 +52,7 @@ class FlashHelper extends Helper
         }
 
         $flash = $options + $flash;
-        $flash['params'] += $this->_config;
+        $flash['params'] = array_merge($flash['params'] , $this->_config);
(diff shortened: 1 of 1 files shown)


ADmad:
Exactly how would one reproduce this error?

makamo:
I just pass my project form cake 3.0 to 3.1

dereuromark:
All keys should be strings, so I doubt that this fixes anything.
Please make an example as requested before.

ADmad:
Value for `params` key is supposed to be always an associative array so the line you are changing should never generate an error. Simply switching from cake 3.0 to 3.1 does not generate any error as you claim.

Closing as the error is most likely being generated due to your app code. If you still the issue is with plugin's code please provide a test case or example code to reproduce the issue.

Issue #42, opened 2015-04-09
Optionally load css and js from CDN in default layout?


bravo-kernel:
I just started testing and I think it would be nice if the default layout has some sort of condition that, when set, downloads the default css and js  from cdn.

This way new users can quickly get up and running with nice visuals without manually having to setup these assets first.

If this sounds reasonable I will create a PR, suggestion on where to set the option would be appreciated.

ADmad:
One issue with using CDN links is we would have keep updating the version number every time there's a new release for a lib. I personally don't want to deal with that.

jadb:
@ADmad that's a very valid point.

@bravo-kernel was suggesting making it optionally turned off. I would suggest making it optionally turned on by defining a CDN URL to use and in the docs on how to install/activate plugin, provide a default one for the ones that don't want to use bower.

Thoughts?

bravo-kernel:
Making it optional would make me just as happy and even though @ADmad has a point I believe the (adoption and quickstart) gains win over impact here:

+ nothing should break if an older lib version is still being used
+ updating two lib references ([css](//maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css) and [javascript](//maxcdn.bootstrapcdn.com/bootstrap/3.3.4/js/bootstrap.min.js)) seems doable compared to e.g. changing layouts

jadb:
How would it be managed? Right now, there are no 'boostrap.php' configuration required and if we could keep it this way, it is better, otherwise it looses from its 'transparency'.

So maybe in the `HelperHelper` configuration?

bravo-kernel:
Closing as I seem to stand alone in this feature but it feels like a missed opportunity from a marketability standpoint (so effectively applied by "other frameworks").

I already imagined the 1-minute video with that much needed "omg, look how easy, let's get that" factor.

To be honest, it just doesn't look as good with the manual editing :wink:

jadb:
You might be right, but it's a thin line between promoting good practices and catering for laziness.

I can accept an alternative layout one could use that could include the hardcoded CDN assets' paths, if that solves it for lazies ;)

bravo-kernel:
Almost entirely duplicating an alternative layout seems a bit non-dry and even more high-maintenance to me but... since this is indeed a very thin line and I personally don't need the turn-key/lazy solution let's not dwell on this any longer.

Maybe somewhere in the future

PR #45, opened 2015-04-13
Adds CdnHelper

bravo-kernel:
This PR is a starting point for solving the feature requested in #42. If the direction is acceptable I will start rounding up. I have chosen a very simple ``CdnHelper``:

+ cdn assets are only used if the helper is loaded
+ one can use the same logic in custom layouts
+ default configuration settings should be both low-maintenance and allow easy user overrides if different URLs are needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ For a complete setup, add the following to your `App\View\AppView`:
 ```php
 public $layout = 'BootstrapUI.default';
 
-public function initialize(array $config)
+public function initialize()
 {
     $this->loadHelper('Html', ['className' => 'BootstrapUI.Html']);
     $this->loadHelper('Form', ['className' => 'BootstrapUI.Form']);
--- a/src/View/Helper/CdnHelper.php
+++ b/src/View/Helper/CdnHelper.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace BootstrapUI\View\Helper;
+
+/**
+ * CndHelper class to insert Bootstrap CDN css and javascript assests into layout.
+ *
+ */
+//class CdnHelper extends \Cake\View\Helper
+class CdnHelper extends \Cake\View\Helper\HtmlHelper
+{
+
+    /**
+     * Default config
+     *
+     * - css: Full URL to most current TwitterBootstrap minified css
+     * - script: Full URL to most current TwitterBootstrap minified js
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'css' => 'http://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css',
+        'script' => 'http://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/js/bootstrap.min.js'
+    ];
+
+    /**
+     *
+     * @return
+     */
+    public function getCss() {
+        return parent::css($this->config('css'));
+    }
+
+    /**
+     *
+     * @return
+     */
+    public function getScript() {
+        return parent::script($this->config('script'));
+    }
+}
--- a/src/View/Helper/CdnHelper.php
+++ b/src/View/Helper/CdnHelper.php
@@ -6,9 +6,15 @@ namespace BootstrapUI\View\Helper;
  * CndHelper class to insert Bootstrap CDN css and javascript assests into layout.
  *
  */
-//class CdnHelper extends \Cake\View\Helper
-class CdnHelper extends \Cake\View\Helper\HtmlHelper
+ class CdnHelper extends \Cake\View\Helper
 {
+
+    /**
+     * @var xxx
+     */
+    public $helpers = [
+        'Html'
+    ];
 
     /**
      * Default config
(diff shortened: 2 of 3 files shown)


src/View/Helper/CdnHelper.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace BootstrapUI\View\Helper;
+
+/**
+ * CndHelper class to insert Bootstrap CDN css and javascript assests into layout.
+ *
+ */
+//class CdnHelper extends \Cake\View\Helper
+class CdnHelper extends \Cake\View\Helper\HtmlHelper
```
ADmad: Don't see the need to extend Html helper.  The helper can use `$helper = ['Htm']` instead.

bravo-kernel: OK, will change

bravo-kernel:
@ADmad thanks for the comment.

If this is an acceptable approach I will add some tests and PHPCS fixes.

src/View/Helper/CdnHelper.php
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * CndHelper class to insert Bootstrap CDN css and javascript assests into layout.
+ *
+ */
+ class CdnHelper extends \Cake\View\Helper
+{
+
+    /**
+     * @var xxx
```
dereuromark: array

bravo-kernel:
Can I take these constructive comments as a <em>yes, this solution is acceptable</em> :wink:

src/View/Helper/CdnHelper.php
```diff
@@ -0,0 +1,49 @@
...
+     * Default config
+     *
+     * - css: Full URL to most current TwitterBootstrap minified css
+     * - script: Full URL to most current TwitterBootstrap minified js
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'css' => 'http://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css',
+        'script' => 'http://maxcdn.bootstrapcdn.com/bootstrap/3.3.4/js/bootstrap.min.js'
```
dereuromark: I wonder if this could be configurable, so you don't have to release new versions after each bugfix release of theirs.

bravo-kernel: Do you mean like e.g. realtime fetching in combination with caching? Feels a bit much to me but hey.

I am thinking about tests using a regex for the version so those stay valid when new releases come out.

dereuromark: No I am neither talking about caching, nor regex testing.
I mean, a configure value to use, with a default set, so people can use different versions if they want or have to.

README.md
```diff
@@ -40,7 +40,7 @@ For a complete setup, add the following to your `App\View\AppView`:
 ```php
 public $layout = 'BootstrapUI.default';
 
-public function initialize(array $config)
+public function initialize()
```
jadb: Good catch! :+1:

src/View/Helper/CdnHelper.php
bravo-kernel: I understood we should stay away from Configure options to keep things as transparent as possible but functional users can already specify their own versions in AppView if these default two don't suit them:

```php
    public function initialize()
    {
        $this->loadHelper('Cdn', [
            'className' => 'BootstrapUI.Cdn',
                'css' => 'http://my.version.',
                'script' => 'http://my.version'
        ]);
    }
```

If I am missing something please let me know.

README.md
bravo-kernel: I rtfd :wink:

src/View/Helper/CdnHelper.php
dereuromark: Fair enough :)

jadb:
@bravo-kernel, short answer, no. But please read as I really appreciate the effort and think I owe an explanation.

While I was looking at the code, I started writing comments. I am copy/pasting the one I was leaving and which kinda turned my opinion around:

> Not a fan of hardcoding it (and supporting it) here. Instead, I would suggest adding instructions for people that wish to use the `CdnHelper` on what lines need to be changed.

This was on the layout file changes. 

When I wrote this, I realized that all this CDN thing should only be in the docs. A lazy person who will have to read the docs to know of the CdnHelper would only have to copy 2 lines and replace the ones included by default with the template.

For maintainers, this would mean only having to update that version part of the paths when new releases are out and would keep the code base lighter (not that it was adding a lot) but most importantly, a good reference to send users to when asking about how to do stuff with CakePHP. Adding bloat and such trivial functionality in a helper promoted by FriendsOfCake won't help.

All that been said, I still hope you'll make the changes to the README to help the lazy use the CDN solution :wink:

bravo-kernel:
ok, fair enough, no problem, at least I learned something about Helpers along the way :smiley:

src/View/Helper/CdnHelper.php
jadb: Heh - reading this now, I had the same comment on that one. I was thinking replacing the `3.3.4` by a token `%s` and add a new config key `version` which would replace the token in `getCss()` and `getScript()`.

jadb: My suggestion was:

```php
// in CdnHelper
protected $_defaultConfig = [
    'css' => 'http://..../%s/....css',
    'script' => 'http://..../%s/....js',
    'version' => '3.3.4',
];
```

So one only needs to set new version when loading helper ;)

bravo-kernel: Hey, looks even cooler, let's implement that :trollface:

PR #55, opened 2015-05-11
add `$this->Flash->render()` to default layout

ayman-alkom:
I didn't edit examples files, because I didnt use them.
But I think they need also

### DIFF
--- a/src/Template/Layout/default.ctp
+++ b/src/Template/Layout/default.ctp
@@ -38,6 +38,12 @@ if (!$this->fetch('tb_body_start')) {
     echo '<body' . $this->fetch('tb_body_attrs') . '>';
     $this->end();
 }
+if (!$this->fetch('tb_flash')) {
+    $this->start('tb_flash');
+    if (isset($this->Flash))
+        echo $this->Flash->render();
+    $this->end();
+}
 if (!$this->fetch('tb_body_end')) {
     $this->start('tb_body_end');
     echo '</body>';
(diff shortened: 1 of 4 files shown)


src/Template/Layout/default.ctp
```diff
@@ -85,6 +85,7 @@ $this->prepend('script', $this->Html->script(['jquery/jquery', 'bootstrap/bootst
 
     <?php
     echo $this->fetch('tb_body_start');
+    echo $this->Flash->render();
```
jadb: Maybe better moved inside the `tb_body_start` on line 39?

ayman-alkom:
What about the examples templates, does they need Flash statement?
I saw that it is very important for the default, but for others I don't know because I haven't used them yet.

jadb:
@ayman-alkom they definitely could use it. The more I think about it and the more I think that the call to Flash should be included within an IF statement (check if helper is loaded in the first place) and yes, in default and other sample templates.

ayman-alkom:
Hi @jadb 
I have tested the code, let'us say that the `tb_body_start` block is initiated in the actions views, so I think that it is more likely to be outside the `tb_body_start` block.
CakePHP conventions generally think by this way.

Otherwise we have to edit the bake templates too.

jadb:
Ok for `tb_flash`.

ayman-alkom:
It is working well for now, I have tested it in my project.

src/Template/Layout/default.ctp
```diff
@@ -38,6 +38,12 @@ if (!$this->fetch('tb_body_start')) {
     echo '<body' . $this->fetch('tb_body_attrs') . '>';
     $this->end();
 }
+if (!$this->fetch('tb_flash')) {
```
jadb: New line and small docblock like for the others please.

jadb:
Thanks! Can you modify the samples as well?

jadb:
Please add dockblock as commented above (to samples as well). Only missing that before I can merge.

ayman-alkom:
If you want the dockblocks for samples can be shorted as following `\\ .....`

jadb:
It's good. Would have preferred new lines to separate stuff, but it's good. I am merging.

PR #57, opened 2015-05-13
Enhance bake index template actions buttons with Icons

ayman-alkom:
Whats about actions as Icons, just a suggestion.
untitled

### DIFF
--- a/src/Template/Bake/Template/index.ctp
+++ b/src/Template/Bake/Template/index.ctp
@@ -75,9 +75,9 @@ $fields = collection($fields)
             $pk = '$' . $singularVar . '->' . $primaryKey[0];
             %>
             <td class="actions">
-                <?= $this->Html->link(__('View'), ['action' => 'view', <%= $pk %>]) ?>
-                <?= $this->Html->link(__('Edit'), ['action' => 'edit', <%= $pk %>]) ?>
-                <?= $this->Form->postLink(__('Delete'), ['action' => 'delete', <%= $pk %>], ['confirm' => __('Are you sure you want to delete # {0}?', <%= $pk %>)]) ?>
+                <?= $this->Html->link('', ['action' => 'view', <%= $pk %>], ['title' => __('View'), 'class' => 'btn btn-default glyphicon glyphicon-eye-open']) ?>
+                <?= $this->Html->link('', ['action' => 'edit', <%= $pk %>], ['title' => __('Edit'), 'class' => 'btn btn-default glyphicon glyphicon-pencil']) ?>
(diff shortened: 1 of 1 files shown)


dereuromark:
If the icons have a title attribute so tooltips are possible that would be great IMO.
Then :+1: from me.

jadb:
I agree with @dereuromark.

ayman-alkom:
That 's a great idea, I tested the title attribute, and the tool-tip didn't triggered automatically, we have to add a piece of code like this:

    $(function () {
      $('[data-toggle="tooltip"]').tooltip()
    })

I appended that code to the script block at the end of index.ctp

dereuromark:
I dont think you need to add that snippet. People can do that on their own (or use any other JS lib). but the title attribute should at least be filled :) And so should the alt tag to be "correct".

src/Template/Bake/Template/index.ctp
```diff
@@ -75,9 +75,9 @@ $fields = collection($fields)
             $pk = '$' . $singularVar . '->' . $primaryKey[0];
             %>
             <td class="actions">
-                <?= $this->Html->link(__('View'), ['action' => 'view', <%= $pk %>]) ?>
-                <?= $this->Html->link(__('Edit'), ['action' => 'edit', <%= $pk %>]) ?>
-                <?= $this->Form->postLink(__('Delete'), ['action' => 'delete', <%= $pk %>], ['confirm' => __('Are you sure you want to delete # {0}?', <%= $pk %>)]) ?>
+                <?= $this->Html->link('', ['action' => 'view', <%= $pk %>], ['title' => __('View'), 'alt' => __('View'), 'class' => 'btn btn-default glyphicon glyphicon-eye-open']) ?>
```
ADmad: `alt` attribute is valid only for `img` tag not `a` tag.

dereuromark: Right, those are not img, but font icons using a/span/i tags. So alt is not necessary after all. In case the font cannot be loaded, what is the "alt" info displayed then?

ADmad: If font icon is not loaded no alternate text will be displayed. Having `title` attribute for `a` tag should trigger tooltip.

@ayman-alkom In which browser(s) did you check whether tooltip is displayed or not?

ayman-alkom: @ADmad In Chrome only

ayman-alkom: I think that things like the compatibility are treated well in the bootstrap it self, and we just have to play the bootstrap well.

ayman-alkom:
sorry for these unnecessarily commits, the final case is without alt tag.
any way I think that we don't have to test our code over the browsers because it's already tested by the bootstrap itself.

ADmad:
Please squash the commits into a single commit.

ayman-alkom:
I have done @ADmad , I am new to contributing but it is going well :+1:

dereuromark:
@ayman-alkom Indeed! :+1:

jadb:
Thanks all! ;)

PR #61, opened 2015-06-08
Correct the appearance  of pagination bar.

ayman-alkom:
screencapture-kobitmobil-com-apps-maras-contacts-1433752936629

To be:

screencapture-kobitmobil-com-apps-maras-contacts-1433754482502

### DIFF
--- a/src/Template/Bake/Template/index.ctp
+++ b/src/Template/Bake/Template/index.ctp
@@ -87,7 +87,7 @@ $fields = collection($fields)
 <div class="paginator">
     <ul class="pagination">
         <?= $this->Paginator->prev('< ' . __('previous')) ?>
-        <?= $this->Paginator->numbers() ?>
+        <?= $this->Paginator->numbers(['before' => '', 'after' => '']) ?>
         <?= $this->Paginator->next(__('next') . ' >') ?>
     </ul>
     <p><?= $this->Paginator->counter() ?></p>


jadb:
@ayman-alkom thanks!

I agree that by default it should look like you're suggesting. Maybe we could add the `after` and `before` keys to the docblock so people know about them?

ADmad:
@ayman-alkom You need to update the tests to match code changes.

@jadb We should remove the default UL wrapping and just get rid of `'before'` and `'after'`. Keeps things as close to core as possible means less possibility of confusion for users.

src/Template/Bake/Template/index.ctp
```diff
@@ -84,11 +84,11 @@ $fields = collection($fields)
         <?php endforeach; ?>
     </tbody>
 </table>
-<div class="paginator">
+<div class="paginator text-center">
```
ADmad: I am :-1: for having the pagination links centered.

ayman-alkom:
@ADmad I am so sorry for the testing, I don't know how to run it.
I like to learn about, I may find the way.
If you have a magic command line I will be happy to know from you.
I have installed the PHPUnit, and ran it from the plugin folder, like this:
"../../bin/phpunit"
and some fatal errors appeared, like this:
`Fatal error: Call to undefined method BootstrapUI\Test\TestCase\View\Helper\FlashHelperTest::hasExpectationOnOutput() `

Is this the right way before continue?

ayman-alkom:
@ADmad  @jadb 
Up to getting rid of 'before' and 'after' keys, I saw that these keys coming as a wrapper about  just the numbers without of the next and before items.
this maybe a mistake in the helper itself, or it is designed in this way.
the old appearance is because we have a two nested UL, one from the template and the second from the keys.

jadb:
Right! Forgot about the template I guess.

@ayman-alkom revert these changes please and just remove it from the template so one does not need to manually add them? That should resolve the issue shown in the screenshot without altering the helper, which otherwise would not be BC anyways.

@ADmad but UL wrapping (AFAIR) is required by bootstrap's pagination?

ADmad:
@jadb Yes it is but let users add that to the template.

jadb:
Hmm, ok I guess but then this PR should point to 0.4 instead of master (because of non-BC change).

On another note @ayman-alkom seemed to have ran into problems trying to drop the keys, but am afraid I did not understand his comment very well.

ADmad:
I don't understand that statement either.

To summarize, changes should be done to both master and 0.4.  

- On master the wrapping UL should be removed from bake template as the PaginatorHelper::numbers() already adds it by default.
- In 0.4 we keep the wrapping UL in template but remove the 'before' and 'after' options in PaginatorHelper::numbers() method.

ayman-alkom:
sorry to be late @jadb and @ADmad 
what I wanted to say ma be explained in this picture:

untitled

so, we can't delete the wrapper ul from the template, because the after and before comes around the numbers li's and the next and before will bi outside of the ul.

I don't if this is a mistake in the original pagination helper it self, or this is the convention.

sorry if my English is not well enough sometimes.

ADmad:
In that case only change that should be done is update the `Paginator::numbers()` call in template to:

```
<?= $this->Paginator->numbers(['before' => '', 'after' => '']) ?>
```

No change is required to helper.

We can change the 'before' & 'after' options in next minor release so that the defaults are similar to the core helper.

PR #64, opened 2015-06-14
radio bugfixes &  improvements

steefaan:
Don't care about the test cases yet, I will write them when you're fine with the changes.

### Before the changes

(1)
```
echo $this->Form->input('radio', [
    'type' => 'radio',
    'options' => [
        'First Radio', 'Second Radio', 'Third Radio'
    ]
]);
```
image

- It was not possible to show the label, even with `label => true`

image

(2)
```
echo $this->Form->input('radio', [
    'type' => 'radio',
    'inline' => true,
    'options' => [
        'First Radio', 'Second Radio', 'Third Radio'
    ]
]);
```
image

- The radios are not aligned correctly
- No container at all

image

(3) With `form-horizontal`
```
echo $this->Form->input('radio', [
    'type' => 'radio',
    'options' => [
        'First Radio', 'Second Radio', 'Third Radio'
    ]
]);
```
image

- As you can see, completly broken

image

### After changes

(1)
image

- Label is customizable, you can event switch off with `label => false`

(2)
image

- Alignment is correct
- Label is customizable, you can even switch off with `label => false`

(3) With `form-horizontal`

(3.1) Default aligment
image

(3.2) Inline aligment
image

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -30,7 +30,8 @@ class FormHelper extends Helper
         'inputContainer' => '<div class="form-group{{required}}">{{content}}{{help}}</div>',
         'inputContainerError' => '<div class="form-group{{required}} has-error">{{content}}{{error}}{{help}}</div>',
         'checkboxWrapper' => '<div class="checkbox"><label>{{input}}{{label}}</label></div>',
-        'radioFormGroup' => '{{label}}<div class="radio-inline-wrapper">{{input}}</div>',
+        'radioInlineFormGroup' => '{{label}}<div class="radio-inline-wrapper">{{input}}</div>',
+        'radioNestingLabel' => '<div class="radio">{{hidden}}<label{{attrs}}>{{input}}{{text}}</label></div>',
         'staticControl' => '<p class="form-control-static">{{content}}</p>'
     ];
 
@@ -170,14 +171,16 @@ class FormHelper extends Helper
                 break;
             case 'radio':
                 $isInline = (isset($options['inline']) && $options['inline'] === true);
-                if (!$isInline || $this->_align === 'horizontal') {
-                    $this->templater()->remove('radioFormGroup');
+
+                $templates = [];
+                if ($isInline && $this->_align !== 'horizontal') {
+                    $templates['formGroup'] = $this->templater()->get('radioInlineFormGroup');
                 }
                 if (!$isInline) {
-                    $this->templater()->add([
-                        'nestingLabel' => '<div class="radio">' . $this->templater()->get('nestingLabel') . '</div>'
-                    ]);
+                    $templates['nestingLabel'] = $this->templater()->get('radioNestingLabel');
                 }
+
+                $this->templater()->add($templates);
                 break;
             case 'select':
                 if (isset($options['multiple']) && $options['multiple'] === 'checkbox') {
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -30,6 +30,7 @@ class FormHelper extends Helper
         'inputContainer' => '<div class="form-group{{required}}">{{content}}{{help}}</div>',
         'inputContainerError' => '<div class="form-group{{required}} has-error">{{content}}{{error}}{{help}}</div>',
         'checkboxWrapper' => '<div class="checkbox"><label>{{input}}{{label}}</label></div>',
+        'multipleCheckboxWrapper' => '<div class="checkbox">{{label}}</div>',
         'radioInlineFormGroup' => '{{label}}<div class="radio-inline-wrapper">{{input}}</div>',
         'radioNestingLabel' => '<div class="radio">{{hidden}}<label{{attrs}}>{{input}}{{text}}</label></div>',
         'staticControl' => '<p class="form-control-static">{{content}}</p>'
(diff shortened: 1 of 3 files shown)


burzum:
:+1:

dereuromark:
Nice one! :)

src/View/Widget/RadioWidget.php
```diff
@@ -32,20 +41,49 @@ public function render(array $data, ContextInterface $context)
             'inline' => false,
         ];
 
+        $templates = $this->_templates->get();
+        $customize = [];
+
         if ($data['inline']) {
-            $this->_templates->add(['radioContainer' => '{{content}}']);
+            $this->_inline = $data['inline'];
+            if ($templates['formGroup'] === '{{label}}{{input}}') {
```
ADmad: Using string matching is flaky. It will break if someone for example adds a space in between label and input.

src/View/Widget/RadioWidget.php
```diff
@@ -32,20 +41,49 @@ public function render(array $data, ContextInterface $context)
...
+                $this->_templates->add(['radioFormGroup' => '{{label}}<div class="radio-inline-wrapper">{{input}}</div>']);
```
ADmad: Hardcoding template here means users can't customize it. Instead the template should be set in `FormHelper::$_templateSet` and switched in appropriate FormHelper method based on options.

src/View/Widget/RadioWidget.php
```diff
@@ -32,20 +41,49 @@ public function render(array $data, ContextInterface $context)
...
+            }
         }
 
-
-        $templates = $this->_templates->get();
-        $customize = [];
-        if ($templates['nestingLabel'] === '{{hidden}}<label{{attrs}}>{{input}}{{text}}</label>') {
+        if (!$data['inline'] && $templates['nestingLabel'] === '{{hidden}}<label{{attrs}}>{{input}}{{text}}</label>') {
```
ADmad: Same comment as above. This check will fail if the template is modified by user.

src/View/Widget/RadioWidget.php
```diff
@@ -32,20 +41,49 @@ public function render(array $data, ContextInterface $context)
...
         unset($data['inline']);
 
         return parent::render($data, $context);
     }
+
+    /**
+     * Renders a label element for a given radio button.
+     *
+     * In the future this might be refactored into a separate widget as other
+     * input types (multi-checkboxes) will also need labels generated.
```
ADmad: Using separate widget would be nice.

steefaan:
@ADmad I will do the test cases as soon you're fine with the changes.

src/View/Widget/RadioWidget.php
steefaan: Let me do this with another PR.

src/View/Helper/FormHelper.php
```diff
@@ -169,7 +167,17 @@ public function input($fieldName, array $options = [])
...
-
+                break;
+            case 'radio':
+                $isInline = (isset($options['inline']) && $options['inline'] === true);
+                if (!$isInline || $this->_align === 'horizontal') {
+                    $this->templater()->remove('radioFormGroup');
+                }
+                if (!$isInline) {
+                    $this->templater()->add([
+                        'nestingLabel' => '<div class="radio">' . $this->templater()->get('nestingLabel') . '</div>'
```
ADmad: There should no hardcoded html.

steefaan: I would keep this otherwise it's more complicate to handle the `nestingLabel`. Wrap the `nestingLabel` in `<div class="radio"></div>` is a special case for radios here. If you place the the `netstingLabel` to the templates directly you need to change the `nestingLabel` for all other inputs conditionally in the helper. Besides the wrapper is forced by bootstrap so I think there is no reason for the user to change the template.

The `nestingLabel` is also customizable this way.

steefaan: @ADmad What do you think?

ADmad: > Besides the wrapper is forced by bootstrap so I think there is no reason for the user to change the template.

Sorry I don't buy that argument. The whole point of the template system is to allow people the ability to customize.

steefaan:
@ADmad Now all the templates are customizable. If you're fine with the changes I will write the test cases now.

ADmad:
@steefaan Templaty sweetness!  I noticed there's a pre-existing hardcoded template on line 187 for 'select' also, could you please fix that too.

steefaan:
Sure, I will do that.

steefaan:
@ADmad I moved the checkbox template and Travis is also fine with the changes now.

jadb:
@steefaan very nice work. 

Could this maybe target `0.4` instead of `master`? 

And just out of curiosity, could `checkbox` on line 164 benefit of the `radio` improvements?

steefaan:
@jadb I will take a look into it.

I think `checkbox` on line 164 can be removed because a single checkbox is always aligned as inline and don't require an additional label. More interesting is the `multiple => checkbox` part form line 186. It's not possible to align multiple checkbox as inline yet and I think this part can benefit from the radio changes. What do you think?

jadb:
@steefaan go for it!

ADmad:
@steefaan Bump

steefaan:
Sorry for the delay guys I will try to finish this as soon as possible.

ADmad:
@steefaan No problem, was just a reminder :)

ADmad:
@jadb Given the display of radio input is pretty broken on 0.3 merging this to master itself should be fine.

dereuromark:
I agree.

steefaan:
It's also an option. I definitely need inline checkboxes later so I will come back either way.

dereuromark:
+1 for incremental improvements - and getting some progress in here again. I dislike the PRs getting stale because too much is requested/done.
Merge, next changes, merge, ...

ADmad:
Great job @steefaan. You can open PR for inline checkboxes at your convenience.

PR #70, opened 2015-07-05
(feature) display the actions links in the collapsed menu

ayman-alkom:
The actions displayed in the side bar can now be shown in the nav bar menu in small screen sizes when the side bar become hidden.

### DIFF
--- a/src/Template/Bake/Element/form.ctp
+++ b/src/Template/Bake/Element/form.ctp
@@ -9,36 +9,35 @@ $fields = collection($fields)
 %>
 <?php
 $this->extend('../Layout/TwitterBootstrap/dashboard');
-$this->start('tb_sidebar');
+$this->start('tb_actions');
 ?>
-<ul class="nav nav-sidebar">
-    <% if (strpos($action, 'add') === false): %>
-    <li><?=
-            $this->Form->postLink(
-            __('Delete'),
-            ['action' => 'delete', $<%= $singularVar
-            %>-><%= $primaryKey[0] %>],
-            ['confirm' => __('Are you sure you want to delete # {0}?', $<%= $singularVar %>-><%= $primaryKey[0] %>)]
-            )
-            ?></li>
-    <% endif; %>
-    <li><?= $this->Html->link(__('List <%= $pluralHumanName %>'), ['action' => 'index']) ?></li>
-    <%
-    $done = [];
-    foreach ($associations as $type => $data) {
-        foreach ($data as $alias => $details) {
-            if ($details['controller'] != $this->name && !in_array($details['controller'], $done)) {
-                %>
-    <li><?= $this->Html->link(__('List <%= $this->_pluralHumanName($alias) %>'), ['controller' => '<%= $details['controller'] %>', 'action' => 'index']) %> </li>
-                <li><?= $this->Html->link(__('New <%= $this->_singularHumanName($alias) %>'), ['controller' => '<%= $details['controller'] %>', 'action' => 'add']) %> </li>
-                <%
-                $done[] = $details['controller'];
-            }
+<% if (strpos($action, 'add') === false): %>
+<li><?=
+    $this->Form->postLink(
+    __('Delete'),
+    ['action' => 'delete', $<%= $singularVar
+    %>-><%= $primaryKey[0] %>],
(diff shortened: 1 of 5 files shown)


src/Template/Bake/Element/form.ctp
```diff
@@ -11,33 +11,31 @@ $fields = collection($fields)
...
-                %>
-    <li><?= $this->Html->link(__('List <%= $this->_pluralHumanName($alias) %>'), ['controller' => '<%= $details['controller'] %>', 'action' => 'index']) %> </li>
-                <li><?= $this->Html->link(__('New <%= $this->_singularHumanName($alias) %>'), ['controller' => '<%= $details['controller'] %>', 'action' => 'add']) %> </li>
-                <%
-                $done[] = $details['controller'];
-            }
+<% if (strpos($action, 'add') === false): %>
+<li><?=
+        $this->Form->postLink(
+        __('Delete'),
```
jadb: missing one indentation level

jadb:
:+1:

ayman-alkom:
what do you think about rename `tb_sidebar` to `tb_actions` or other suitable one

jadb:
Renaming, maybe not as it would break for others. Deprecating by using an alias now  and completely removing it later on, isn't something I'll oppose. @ADmad thoughts?

ADmad:
Don't see any real need to rename. People could add other things to the sidebar too if they want. I actually preferred having the <ul> tags in each element.

jadb:
@ayman-alkom how about by default `tb_actions` would internally re-use `tb_sidebar` and we move the `<ul>` tag back into the element. This way, one could customize to his hearts content without having to re-write the default template. What do you think?

/cc @ADmad

ayman-alkom:
We can do one of the following:
 1. create the `tb-actions` from the li's only, then:
	 - wrapping `tb-actions` by UL to make the `tb_side_bar`.
	 - and for the nav, we can do the same and get `tb_nav_actions` or even just wrapping `tb_actions` immediately in the nav without a new Block
 2. As the deffirence between `tb_side_bar` and `tb_nav_action` is just the UL's class, I dont know if there is a way to use the same block and passing the `class_string` as a parameter.

ayman-alkom:
Hi @jadb , I hope it has been done as you explained

jadb:
@ayman-alkom I am re-opening if you don't mind as I was planning on toying it with over the weekend (hopefully!) before merging it.

@ADmad have you checked the latest changes? Opinion?

src/View/Helper/FormHelper.php
```diff
@@ -134,6 +134,12 @@ public function create($model = null, array $options = [])
         return parent::create($model, $this->_formAlignment($options));
     }
 
+    public function submit($caption = null, array $options = [])
```
ADmad: API docblock is missing.

ayman-alkom:
I am sorry to close it, but found my self adding more commits out of the title of the PR, and I thought this PR will not applied.
if you want I can delete the last commits, to redo them again in other PR

ADmad:
@jadb Now that crud-view is in pretty good shape i wonder if it's worth maintaining bake templates in this plugin. We have enough to deal with already :)

jadb:
@ADmad maintaining, I agree - but not everyone uses CrudView - should it not cater to these people as well? Or maybe we could have a separate plugin (encouraging the community to maintain it) for BUI baking?

ADmad:
> maintaining, I agree - but not everyone uses CrudView - should it not cater to these people as well? 

I don't see anyone using the templates (baked or the example ones) included in the plugin for anything besides admin backends (I could be wrong), in which case they should be just using crud-view IMO :)

> Or maybe we could have a separate plugin (encouraging the community to maintain it) for BUI baking?

Yeah separate plugin would be fine, though frankly I don't much faith in the community to maintain it unless a particular individual steps up and take up the role of maintainer.

jadb:
I agree with your first point @ADmad - however, CrudView requires Crud and not everyone uses that for their backends...

ayman-alkom:
I have delete the out of scope commits, and squashed the rest.
later I can pull a new PR for the deleted commits.

jadb:
Thanks! :+1:

PR #74, opened 2015-07-14
Form helper submit, add `btn` class by default.

ayman-alkom:
I updated the test file, but this may the first time I work with testing :).

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -152,7 +152,7 @@ class FormHelper extends Helper
     {
         $class = ['btn'];
         if (array_key_exists('class', $options)) {
-            $optionsClass = explode(' ', $options['class']);
+            $optionsClass = is_array($options['class']) ? $options['class'] : explode(' ', $options['class']);
             $class = array_unique(array_merge($class, $optionsClass));
         }
         $options = Hash::merge($options, ['class' => $class]);
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -150,7 +150,13 @@ class FormHelper extends Helper
      */
     public function submit($caption = null, array $options = [])
     {
-        $options = Hash::merge($options, ['class' => 'btn']);
+        $class = ['btn'];
+        if (array_key_exists('class', $options)) {
+            $optionsClass = explode(' ', $options['class']);
+            $class = array_unique(array_merge($class, $optionsClass));
+        }
+        $options = Hash::merge($options, ['class' => $class]);
+
         return parent::submit($caption, $options);
     }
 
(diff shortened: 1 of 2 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -135,6 +136,25 @@ public function create($model = null, array $options = [])
...
+     *  extension .jpg, .jpe, .jpeg, .gif, .png use an image if the extension
+     *  exists, AND the first character is /, image is relative to webroot,
+     *  OR if the first character is not /, image is relative to webroot/img.
+     * @param array $options Array of options. See above.
+     * @return string A HTML submit button
+     * @link http://book.cakephp.org/3.0/en/views/helpers/form.html#creating-buttons-and-submit-elements
+     */
+    public function submit($caption = null, array $options = [])
+    {
+        $options = Hash::merge($options,['class' => 'btn']);
```
ADmad: Space missing after `$options,`

ayman-alkom:
the PR is ready now, up to styles I am not sure if they are OK or not.

src/View/Helper/FormHelper.php
```diff
@@ -135,6 +136,31 @@ public function create($model = null, array $options = [])
...
+     *  OR if the first character is not /, image is relative to webroot/img.
+     * @param array $options Array of options. See above.
+     * @return string A HTML submit button
+     * @link http://book.cakephp.org/3.0/en/views/helpers/form.html#creating-buttons-and-submit-elements
+     */
+    public function submit($caption = null, array $options = [])
+    {
+        $class = ['btn'];
+        if (array_key_exists('class', $options)) {
+            $optionsClass = explode(' ', $options['class']);
```
ADmad: You are assuming `class` will be string, it can be array too. So this whole is not necessary imo, if an array is passed 'btn' will be merged and if someone wants to override then can use string.

ADmad: Or you can just use OptionsAwareTrait::injectClasses()

src/View/Helper/FormHelper.php
```diff
@@ -135,6 +136,31 @@ public function create($model = null, array $options = [])
...
+     *  OR if the first character is not /, image is relative to webroot/img.
+     * @param array $options Array of options. See above.
+     * @return string A HTML submit button
+     * @link http://book.cakephp.org/3.0/en/views/helpers/form.html#creating-buttons-and-submit-elements
+     */
+    public function submit($caption = null, array $options = [])
+    {
+        $class = ['btn'];
+        if (array_key_exists('class', $options)) {
+            $optionsClass = is_array($options['class']) ? $options['class'] : explode(' ', $options['class']);
```
jadb: How about changing that to:

```
$optionsClass = explode(' ', (array)$options['class']);
```

jadb:
Thanks! LGTM.

PR #77, opened 2015-07-21
Remove href from disabled pagination links

deizel:
On the first or last page of a paginated action, the disabled next/prev links were still clickable.

This change doesn't seem to affect styling. Also, having an empty `href` attribute is against Yslow's rules.

### DIFF
--- a/src/View/Helper/PaginatorHelper.php
+++ b/src/View/Helper/PaginatorHelper.php
@@ -17,10 +17,10 @@ class PaginatorHelper extends \Cake\View\Helper\PaginatorHelper
         $this->_defaultConfig['templates'] = [
             'nextActive' => '<li class="next"><a rel="next" aria-label="Next" href="{{url}}">' .
                             '<span aria-hidden="true">{{text}}</span></a></li>',
-            'nextDisabled' => '<li class="next disabled"><a href=""><span aria-hidden="true">{{text}}</span></a></li>',
+            'nextDisabled' => '<li class="next disabled"><a><span aria-hidden="true">{{text}}</span></a></li>',
             'prevActive' => '<li class="prev"><a rel="prev" aria-label="Previous" href="{{url}}">' .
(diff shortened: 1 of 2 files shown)


ADmad:
Are you sure removing the href doesn't cause styling issues with bootstrap?

deizel:
Styling is the same in both Chrome/Firefox, including the disabled cursor when you mouseover.

pagination 2015-07-21 22-08-04

dereuromark:
I think we had that discussion in the core, and we ended up adding a small JS snippet to avoid that instead. Styling can indeed be affected with a few older browsers.

ADmad:
Bootstrap itself only supports IE9+ isn't it? Theb we don't need to worry about older browsers.

dereuromark:
I am still in favor of making it consistent to the core version.

ADmad:
That would be pointless since whole idea of this plugin is to customize the default templates to suite bootrap.

@jadb Ping

ayman-alkom:
I am agree with @ADmad

jadb:
I must admit that I think like @dereuromark when it comes to sticking to the way core does it on many fronts. Maybe this should have closely followed the core after his PR there.

However, in situations like this one, where adding support to match the core's support doesn't bring any benefit aside from aligning with core (since as @ADmad said, only [9+](http://getbootstrap.com/getting-started/#support) are supported), I think it's best to skip it. Those templates are to be customized by users, and the least unnecessary bloat in them, the better I think.

ADmad:
> I must admit that I think like dereuromark when it comes to sticking to the way core does it on many fronts.

In general I do too (That's the reason I actually dislike the 'prev', 'next' options added to PaginatorHelper::numbers()). My comment was strictly in reference to this change will relates to templates.

ADmad:
Merging. Just in case the change does cause an issue in modern browsers we can revisit/revert.

PR #80, opened 2015-07-24
Enhance the look of date selects

ayman-alkom:
the look will be like in this pic:
screencapture-www-kobitmobil-com-apps-maras-20categories-add-1437737424805

this PR may need some ideas, like make the column sizes deynamic (calculated with selects numbers in consideration )

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -28,8 +28,8 @@ class FormHelper extends Helper
     protected $_templates = [
         'error' => '<div class="help-block">{{content}}</div>',
         'help' => '<div class="help-block">{{content}}</div>',
-        'inputContainer' => '<div class="form-group{{required}}">{{content}}{{help}}</div>',
-        'inputContainerError' => '<div class="form-group{{required}} has-error">{{content}}{{error}}{{help}}</div>',
+        'inputContainer' => '<div class="form-group {{type}}{{required}}">{{content}}{{help}}</div>',
+        'inputContainerError' => '<div class="form-group {{type}}{{required}} has-error">{{content}}{{error}}{{help}}</div>',
         'checkboxWrapper' => '<div class="checkbox"><label>{{input}}{{label}}</label></div>',
         'multipleCheckboxWrapper' => '<div class="checkbox">{{label}}</div>',
         'radioInlineFormGroup' => '{{label}}<div class="radio-inline-wrapper">{{input}}</div>',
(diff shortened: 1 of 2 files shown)


ayman-alkom:
the selects now look like this 
screencapture-www-kobitmobil-com-apps-maras-20categories-add-1437737424805

dereuromark:
Looks better then the current output for sure :)
Might be worse showing that to compare em

jadb:
I like the result - as @dereuromark suggested, please post screenshot of current look for comparison and if you could update the tests too. @ADmad, thoughts?

ayman-alkom:
The default look was:
untitled-2

This look when using the grid system:
```
    <div class="row">
        <div class="col-xs-3">{{year}}</div>
        <div class="col-xs-3">{{month}}</div>
        <div class="col-xs-2">{{day}}</div>
        <div class="col-xs-2">{{hour}}</div>
        <div class="col-xs-2">{{minute}}</div>
        <div class="col-xs-2">{{second}}</div>
        <div class="col-xs-2">{{meridian}}</div>
    </div>
```
untitled-3

this is the look when using inline list:
```
    <ul class="list-inline">
        <li>{{year}}</li>
        <li>{{month}}</li>
        <li>{{day}}</li>
        <li>{{hour}}</li>
        <li>{{minute}}</li>
        <li>{{second}}</li>
        <li>{{meridian}}</li>
    </ul>
```
untitled-1

ayman-alkom:
we can add classes as the following:
```
<ul class="list-inline">
    <li class'"year">{{year}}</li>
    <li class'"month">{{month}}</li>
    <li class'"day">{{day}}</li>
    <li class'"hour">{{hour}}</li>
    <li class'"minute">{{minute}}</li>
    <li class'"second">{{second}}</li>
    <li class'"meridian">{{meridian}}</li>
</ul>
```
in this case we can control the width of each elm separately

ayman-alkom:
another idea:
we can make these strings [$_templates] customizable  by using configuration keys.

bravo-kernel:
:100:

ayman-alkom:
I have resubmitted the commits and renamed them.

ayman-alkom:
two points now for discus:
   - adding styles to the plugin or leave it to the user (styles that manage the widths in date,dateTime and other cases)
   - make all the string templates configurable, or just the dateWidget, or none (because it can be overrid anyway)

jadb:
@ayman-alkom great work!

- I wouldn't add styles, lots of people already customize their templates, but would add something to the examples.
- Also, I'd skip that for the same reason you are mentioning. I can see how it would help making it even easier but again, an example should be enough for anyone to replicate and less code to maintain :wink: If however, others would like to see it part of the plugin, I won't oppose it but maybe in a separate PR?

ayman-alkom:
Thanks @jadb 
in this case this PR is may finished, and we may add some tutorials pages to produce the ready to use styles, and the way to override the strings templates.

dereuromark:
Awesome! :+1:

jadb:
:+1:

PR #81, opened 2015-07-31
add date.css to examples

ayman-alkom:


lorenzo:
What does this change?

dereuromark:
@ayman-alkom It helps to provide a before/after screenshot comparison.

ayman-alkom:
In fact I think that this PR is not necessary, it is just about setting the widths of the date input select items, in addition to some media query, witch can any one do in his own way.

ayman-alkom:
@dereuromark this the after/before you mentioned
untitled

PR #85, opened 2015-08-22
Enhancements to Flash Helper and Element

darensipes:
Created a way a to not escape message content so thinks like links are not escaped. Works like most of CakePHP helper in that you must set escape to false, otherwise it escapes by default. 

```php
$this->Flash->success('<i class="fa fa-trash"></i> Your tag was trashed!', ['params' => ['escape' => false]]);
```

I also thought that sometimes you might need to do some additional customizations to the Flash Element that is used by default. So it would be nice to set the element when you load the helper.

```php
$this->loadHelper('BootstrapUI.Flash', ['element' => 'Flash/default']);
```

### DIFF
--- a/src/Template/Element/Flash/default.ctp
+++ b/src/Template/Element/Flash/default.ctp
@@ -1,6 +1,6 @@
 <?php
 $class = array_unique((array)$params['class']);
-$message = h($message);
+$message = (isset($params['escape']) && $params['escape'] === false) ? $message : h($message);
 
 if (in_array('alert-dismissible', $class)) {
     $button = <<<BUTTON
--- a/src/View/Helper/FlashHelper.php
+++ b/src/View/Helper/FlashHelper.php
@@ -21,7 +21,8 @@ class FlashHelper extends Helper
      */
     protected $_defaultConfig = [
         'class' => ['alert', 'alert-dismissible', 'fade', 'in'],
-        'attributes' => ['role' => 'alert']
+        'attributes' => ['role' => 'alert'],
+        'element' => 'BootstrapUI.Flash/default'
     ];
 
     /**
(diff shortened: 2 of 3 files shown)


src/Template/Element/Flash/default.ctp
```diff
@@ -10,4 +10,3 @@ BUTTON;
 }
 
 echo $this->Html->div($class, $message, $params['attributes']);
-?>
```
ADmad: Please add back the closing `?>`. It's omitted only for class files not templates.

darensipes: No Problem. I added it back. Thanks for catching my oversight!

ADmad:
Looks good to me. @jadb ?

jadb:
LGTM :smile: 

Could you please squash the commits @darensipes ?

src/Template/Element/Flash/default.ctp
```diff
@@ -1,6 +1,6 @@
 <?php
 $class = array_unique((array)$params['class']);
-$message = h($message);
+$message = isset($params['escape']) && $params['escape'] === false ? $message : h($message);
```
jadb: Shouldn't the condition check be in `()`? This will fail and return no message if `escape` is not set - https://3v4l.org/R6jU1

darensipes:
Commit Squashed! What were you thinking for the test? One for escape param?

src/Template/Element/Flash/default.ctp
darensipes: No Problem! Thanks for checking. You had me second guessing myself for a moment, I was thinking of the thousands of ternary statements I have used in the past that looked just like that!

darensipes:
I added one test. Let me know if that is sufficient.

jadb:
:+1:

PR #88, opened 2015-08-29
Fixed Readme mistakes I made

Modicrumb:
Readme now has intended instructions. This is dealing with the directory structure changes I proposed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ A quick way of getting the Bootstrap assets installed is using [bower]. Assuming
 
 ```
 bower install bootstrap
-mkdir -p webroot/bootstrap
+mkdir -p webroot/bootstrap webroot/bootstrap/css webroot/bootstrap/js webroot/bootstrap/jquery webroot/bootstrap/fonts
 cp bower_components/bootstrap/dist/css/* webroot/bootstrap/css/.
 cp bower_components/bootstrap/dist/js/* webroot/bootstrap/js/.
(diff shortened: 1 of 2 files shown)


davidyell:
I think you should be suggesting using a symlink. Especially with Bower, as if I update my dependancy, I don't want to have to copy all the files again, I'd rather it just update automatically.

Modicrumb:
Sounds like a better solution, my first idea was to create a gulp file that would move and update things automatically, but maybe a .bowerrc file would be better? http://stackoverflow.com/questions/19251931/symlink-or-copy-files-out-of-bower-component. Truthfully I am not too familiar with the idea of symlinking JavaScript/CSS assets. Is this done at the apache level? Got a link where I can read about it?

ADmad:
@Modicrumb Just standard *nix symlinks (using `ln -s`).

Modicrumb:
Yeah I think I will close this, it really isn't a great solution. Just the only one I thought of and stuck to. I do like the idea of keeping bootstrap assets in one folder in the pull request. The folder structure is a bit awkward with the fonts because they have to be put in the css folder for the original instructions to work. I can look into .bowerrc and symlinking instead. I feel it isn't an issue to warrant a lot of attention looking back at it since the brunt of the plug in is about facilitating writing mark up that is bootstrap compatible and people can set the layout up however they please. Thanks for the feedback.

PR #93, opened 2015-09-13
Fixed indentation in the generated view's.

martonmiklos:
The bake generated views indentation was ugly. 
Modified the add/edit forms submit button text to Save for edit and Add for the add.


jadb:
Thanks! :+1:

PR #99, opened 2015-10-29
Flash helper edits so it can accept multiple flashes

Modicrumb:


ADmad:
Support for multiple flashes is already available on `master` branch. 0.4 is a dev branch which is lagging behind master, you shouldn't be using it.

ADmad:
Please open a PR against master branch if you think it has issue handing multiple flashes.

dereuromark:
@ADmad I dont see it working yet for the latest release, though.

ADmad:
Like I said a PR for master is welcome.

dereuromark:
@Modicrumb Yes please please :)

dereuromark:
Ah, I think a new release is missing
I am using tagged releases, currently `0.3.2` (not dev-master).
So please release a new version and then it hopefully all works out.

ADmad:
It might be outdated but it's a feature branch which could be picked up again in the future. You should use `~3.0` as the master branch readme states.

ADmad:
It might be outdated but it's a feature branch which could be picked up again in the future. You should use `~3.0` as the master branch readme states.

PR #100, opened 2015-11-02
Unset class to avoid re-merging.

dereuromark:
Without removing the array key, the result of Hash::merge() will contain the provided classes multiple times.

    <div class="submit"><input type="submit" class="btn-primary btn btn-primary" value="Foo"/></div>


ADmad:
A test would be nice.

dereuromark:
Done

PR #102, opened 2015-12-02
Add configuration where necessary to follow FriendsOfCake standards

josegonzalez:


josegonzalez:
Some of these changes are maybe suspect, but I figure they haven't hurt on other projects I've worked on, so I don't think anyone will mind.

We really do need to cleanup the travis:3.x branch and start pushing it's usage out everywhere...

.travis.yml
```diff
@@ -6,34 +6,55 @@ php:
...
     - DEFAULT=1
 
 matrix:
+  allow_failures:
+    - php: hhvm
+
   fast_finish: true
 
   include:
-  - php: 7.0
```
lorenzo: No php 7?

lorenzo: Better said... why use a slower version of php for checking coding standards?

josegonzalez: You could say the same about coveralls, no?

josegonzalez: At the end of the day, the php version for coding standards shouldn't matter imo.

.coveralls.yml
```diff
@@ -0,0 +1,3 @@
+src_dir: .
```
ADmad: Coverage reporting on coveralls.io seems to works just fine with this. I don't really mind adding this file but would prefer if we don't add extra meta data files if not really needed.

.travis.yml
ADmad: Using php 7 helps see build results faster :)  That's enough incentive for me to use php 7.

lorenzo: Exactly the version does not matter there, it is just a lot faster. Unfortunately the php version usually does matter for the coverage.

.coveralls.yml
josegonzalez: If that's the case, I'd also like to get rid of the `.semver` file we seem to add everywhere.

lorenzo: :+1:

dereuromark: :+1:

.travis.yml
josegonzalez: So coveralls breaks under php 7 but phpcs is like "yeah sure why not".

Computers are the fucking worst.

dereuromark:
:+1:

composer.json
```diff
@@ -14,20 +10,25 @@
...
             "homepage": "https://github.com/friendsofcake/bootstrap-ui/graphs/contributors"
         }
     ],
-    "support": {
-        "issues": "http://github.com/friendsofcake/bootstrap-ui/issues",
-        "source": "http://github.com/friendsofcake/bootstrap-ui"
+    "require": {
+        "cakephp/cakephp": "~3.0",
+        "php": ">=5.4.0",
+        "cakephp/plugin-installer": "*"
```
ADmad: This is not needed. Plugin installer is required by app.

PR #104, opened 2015-12-28
Added support for templateVars added in 3.1.0

redd:


ADmad:
Can you add tests for the change?

redd:
I think it's not part of plugin, so it's already tested in framework.

PR #106, opened 2016-01-05
Grid support

golgher:
Added methods to generate Bootstrap grid.

Code example:
```php
echo $this->Html->gridContent($this->Form->input('field1'))
    ->gridConfig(['size' => 4])
    ->gridContent($this->Form->input('field2'))
    ->gridConfig(['size' => 4])
    ->gridContent($this->Form->input('field3'))
    ->gridConfig(['size' => 4])
    ->gridRender();
```

### DIFF
--- a/src/View/Helper/HtmlHelper.php
+++ b/src/View/Helper/HtmlHelper.php
@@ -5,6 +5,12 @@ class HtmlHelper extends \Cake\View\Helper\HtmlHelper
 {
     use OptionsAwareTrait;
 
+    private $grid = [];
+
+    private $gridOptions = [];
+
+    private $gridCounter = 0;
+
     /**
      * Returns Bootstrap badge markup. By default, uses `<SPAN>`.
      *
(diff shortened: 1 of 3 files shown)


src/View/Helper/HtmlHelper.php
```diff
@@ -5,6 +5,12 @@ class HtmlHelper extends \Cake\View\Helper\HtmlHelper
 {
     use OptionsAwareTrait;
 
+    private $grid = [];
+
+    private $gridOptions = [];
+
+    private $gridCounter = 0;
```
dereuromark: Does it have to be private? Often times people extend this HtmlHelper and would have to redeclare private stuff.
Protected should suffice IMO.

dakota:
I like the idea, but the interface looks very alien and "uncakelike". I'd rather see something like `tableCells`.

dereuromark:
Wouldnt this maybe be an own GridHelper instead of putting all that into the Html one?

src/View/Helper/HtmlHelper.php
dakota: I would say that the FoC plugins should follow the same standards of core CakePHP where no private methods or properties are allowed.

ADmad:
- Don't use `private` properties or methods
- Don't uses HTML fragments like in `gridRender()`. Either use string templates and/or elements.
- As @dereuromark sadi better to have a separate helper instead of adding more methods to HtmlHelper

src/View/Helper/HtmlHelper.php
jippi: :+1:

golgher:
I agree with almost everything, I just think it should be part of the Html Helper because it will render html code, but that's my opinion.
I think the project could implement other Bootstrap's components inside the HtmlHelper, such as: Navs (tabs and pills), List Group, Panels, Wells, Grid, etc.

dereuromark:
Almost every helper generates HTML :) See Form or other helpers.
The question is more: Is it a large enough subpart to be its own helper or is it OK to through into the existing one.
If you say "Navs (tabs and pills), List Group, Panels, Wells, Grid, etc" then this sounds to me like a new separate one rather than bloating the HtmlHelper IMO.

golgher:
Sorry, I think I explained wrong. I think the HtmlHelper of the Bootstrap's project, should know how to write the html components of Bootstrap.

lorenzo:
@golgher Having it there makes sense to me

jadb:
I don't think the `HtmlHelper` should be bloated with `grid` nor `components`. I started things in here because it was simpler but if there is a need to add other features, they are complex domains that would require their own respective helpers IMO.

ADmad:
Closing this due to lack of activity. 

@golgher You are welcome to reopen or make a new PR with changes as requested.

PR #107, opened 2016-01-09
HTML5 shim and Respond.js update

balcsida:
Updated to the latest HTML5 shim an Respond.js
Added some extra ident. to make HTML output prettier


ADmad:
I would rather we get rid of this cruft. It's 2016!  Those unfortunate ones who still need to supported outdated IEs can use their own templates.

lorenzo:
I think we can discuss removing it it in a following pull request. This is good enough :)

PR #108, opened 2016-01-12
add document variable @var to templates, representing View class

ayman-alkom:
I found this helps with the editors auto complete

### DIFF
--- a/src/Template/Bake/Element/form.ctp
+++ b/src/Template/Bake/Element/form.ctp
@@ -8,6 +8,7 @@ $fields = collection($fields)
 });
 %>
 <?php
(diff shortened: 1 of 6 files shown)


ADmad:
@jadb Are you okay with this? Personally not to happy having it in each file but meh.

jadb:
@ADmad, it seems to help people on IDEs. They got added on core (specifically bake), so let's follow I say.

ADmad:
Okay, then.

ayman-alkom:
I have another pull request, and wait for this to be merged, I made a new branch but I had some problems then.

PR #110, opened 2016-01-12
Add UIView class

ayman-alkom:
I thought that having this class will make the setup more simple and the code more clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,17 +42,48 @@ or using CakePHP's console:
 ./bin/cake plugin load BootstrapUI
 ```
 
-For a complete setup, add the following to your `src\View\AppView`:
+## AppView Setup
+Then for a complete setup, just make your `AppView` class extends `BootstrapUI\View\UIView`.
+The `src\View\AppView.php` then will be as the following: 
 
 ```php
-public $layout = 'BootstrapUI.default';
+namespace App\View;
 
-public function initialize()
+use BootstrapUI\View\UIView;
+
+class AppView extends UIView
 {
-    $this->loadHelper('Html', ['className' => 'BootstrapUI.Html']);
-    $this->loadHelper('Form', ['className' => 'BootstrapUI.Form']);
-    $this->loadHelper('Flash', ['className' => 'BootstrapUI.Flash']);
-    $this->loadHelper('Paginator', ['className' => 'BootstrapUI.Paginator']);
+
+    /**
+     * Initialization hook method.
+     */
+    public function initialize()
+    {
+        //Don't forget to call the parent::initialize()
+        parent::initialize();
+    }
+}
+```
+
+### AppView Setup Using UIViewTrait
+```php
+namespace App\View;
+
+use Cake\View\View;
+
+class AppView extends View
+{
+
+    use UIViewTrait;
+    
+    /**
+     * Initialization hook method.
+     */
+    public function initialize()
+    {
+        //render the initializeUI method from the UIViewTrait
+        $this->initializeUI();
+    }
 }
 ```
 
--- a/src/View/UIView.php
+++ b/src/View/UIView.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace BootstrapUI\View;
+
+use Cake\View\View;
+
+/**
+ * UIView, the customised BootstrapUI View class.
+ *
+ * customises the $layout to the BootstrapUI's layout
+ * and initialize the View with the BootstrapUI's helpers
+ */
+class UIView extends View
+{
+    use UIViewTrait;
+
+    /**
+     * Initialization hook method.
+     *
+     * @return void
+     */
+    public function initialize()
+    {
+        parent::initialize();
+
+        //render the initializeUI method from the UIViewTrait
+        $this->initializeUI();
+    }
+}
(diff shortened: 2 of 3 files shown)


ayman-alkom:
For a complete setup now, as described in the READ.ME, We go to `src\View\AppView`:
- replace `use Cake\View\View;` with `use BootstrapUI\View\UIView as View;`
- add `parent::initialize();` to the `initialize` function.
```php
namespace App\View;

use BootstrapUI\View\UIView as View;

class AppView extends View
{

    public function initialize()
    {
        parent::initialize();
    }
}
```

renan:
I would rather have it without the alias. `class AppView extends UIView` should be enough

ADmad:
Yup the aliasing is not necessary.

src/View/UIView.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Cake\View\View;
+
+/**
+ * BootstrapUI’s  view class
+ */
+class UIView extends View
+{
+
+    public $layout = 'BootstrapUI.default';
```
ADmad: Docblock missing. Wonder by the phpcs build did not report it.

ayman-alkom:
I may made the aliasing for the new programmers, making the setup simple as possible.
But I agree that there is no need to it.
I will update the read.me anyway

src/View/UIView.php
ayman-alkom: Some thing like this works?
````php
/**
 * UIView, the extended class of View one. UIView customise the $layout to the BootstrapUI's layout
 * and initialize the View with the BootstrapUI's helpers.
 *
 * To use UIView class just extend it and make sure to render parent::initialize();
 * The full example here:
 *
 * ```
 * namespace App\View;
 * 
 * use BootstrapUI\View\UIView as View;
 * 
 * class AppView extends View
 * {
 *      public function initialize()
 *      {
 *          parent::initialize();
 *      }
 * }
 * ```
 */
````

**Comment by lorenzo** [comment]
I already have my own view where I do a bunch of stuff. I think it would be better to offer the interesting parts of this new UIView class into a trait. So that it can be reused in other classes.

**Comment by ayman-alkom** [comment]
@lorenzo it seems a good idea to use a treat, but using UIView still more clean for our code, I see, we may create a treat and use it in the UIView whereas we provide two options to setup the app Treat and UIView.

**Comment by lorenzo** [comment]
Having the UIview class and the trait would be good. The view class would then use the trait

**Comment by jadb** [comment]
Sorry for the delay @ayman-alkom. Can you please rebase?

File: README.md [hunk 2]
```diff
@@ -36,17 +36,21 @@ You then need to load the plugin. In `boostrap.php`, something like:
...
 
-public function initialize()
+class AppView extends UIView
 {
-    $this->loadHelper('Html', ['className' => 'BootstrapUI.Html']);
-    $this->loadHelper('Form', ['className' => 'BootstrapUI.Form']);
-    $this->loadHelper('Flash', ['className' => 'BootstrapUI.Flash']);
-    $this->loadHelper('Paginator', ['className' => 'BootstrapUI.Paginator']);
+
+    public function initialize()
```
dereuromark: Would be good to have docblock here, too.

ayman-alkom:
no problem @jadb , it 's been rebased I think

src/View/UIViewTrait.php
```diff
@@ -0,0 +1,31 @@
...
+ * setting the View layout to the UIBootstrap's one.
+ */
+trait UIViewTrait
+{
+    /**
+     * Initialization hook method.
+     *
+     * @return void
+     */
+    public function initialize()
```
ADmad: I wonder if this method should be named something else and the initialize() should be in the UIView instead which calls this method. @lorenzo Thoughts?

ayman-alkom: Yes I agree with you @ADmad , that will be better

README.md
ayman-alkom: yes that will be better

ayman-alkom:
Hi @jadb , @ADmad  any response ?

src/View/UIViewTrait.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * Initialization hook method.
+     *
+     * @return void
+     */
+    public function initializeUI()
+    {
+        //Set the layout to BootstrapUI's layout
+        $this->layout = 'BootstrapUI.default';
```
ADmad: The layout should be set only if current value is `default`, otherwise it will override custom layout passed through controller which is set in View::__construct() before initialize() is called.

ayman-alkom: OK

ayman-alkom: Hi @ADmad , it has done, I added the test for empty value too

src/View/UIViewTrait.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Initialization hook method.
+     *
+     * @return void
+     */
+    public function initializeUI()
+    {
+        if (empty($this->layout) || $this->layout === 'default') {
+            //Set the layout to BootstrapUI's layout
```
ADmad: This inline comment and the one below is not necessary.

ADmad:
@lorenzo Any more feedback on this?

lorenzo:
Looks good :D

Issue #111, opened 2016-01-19
Bake: don't duplicate postLink forms


bfncs:
When using bake with the provided templates, in the resulting templates the forms generated via `FormHelper::postLink()` [are duplicated for usage in the navbar and the sidebar](https://github.com/FriendsOfCake/bootstrap-ui/blob/cce5787b791a81e4901d8a5443a8674fcbe20a83/src/Template/Bake/Template/view.ctp#L65)? These forms [rely on a unique ID](https://github.com/cakephp/cakephp/blob/3.1.7/src/View/Helper/FormHelper.php#L1645) to be successfully posted, which doesn't work any more:

    Uncaught TypeError: document.post_569e5bd99a06c486556933.submit is not a function

It would be nice if you would provide a working default here. The obvious solution would be to simply duplicate the content for the sidebar instead of reassigning them, but this produces a lot of unneeded boilerplate. Maybe there's an even more elegant solution?

josegonzalez:
@bfncs maybe move that section to an element and assign it to a separate block for reuse? Mind making a PR?

bfncs:
Thanks for the quick replay, @josegonzalez.

That's probably better than duplicating the blocks in place, but it would require one element per view, messing up elements – is this really an improvement? Can't think of a better approach currently though.

josegonzalez:
How would it mess up elements?

josegonzalez:
It's that or a foreach loop, take your pick :)

bfncs:
Ok, I'll give it a try…

jadb:
Resolved in #112

PR #112, opened 2016-01-19
Fix duplicate postLink forms in baked templates

bfncs:
Fix for #111

I decided for `foreach` instead of elements. Still ugly, but get's the job done.


jadb:
Thanks!

PR #113, opened 2016-01-20
input templates fix

jrble819:
Fix to allow templates to be passed in for inputs.  Currently there appears to be no way to change templates on the fly for an input.  Cake's Form Helper takes care of this but there was no way to pass the templates to the bootstrap UI and have them be used.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -191,6 +191,14 @@ class FormHelper extends Helper
         $options = $this->_parseOptions($fieldName, $options);
         $reset = $this->templates();
 
+        $templater = $this->templater();
+        if (!empty($options['templates'])) {
+            $templater->push();
+            $templateMethod = is_string($options['templates']) ? 'load' : 'add';
+            $templater->{$templateMethod}($options['templates']);
+        }
+        unset($options['templates']);
+
         switch ($options['type']) {
             case 'checkbox':
                 if (!isset($options['inline'])) {


src/View/Helper/FormHelper.php
```diff
@@ -191,6 +191,14 @@ public function input($fieldName, array $options = [])
         $options = $this->_parseOptions($fieldName, $options);
         $reset = $this->templates();
 
+        $templater = $this->templater();
+        if (!empty($options['templates'])) {
```
dereuromark: Do you need isset()? Otherwise the key+value will stick around in the options array.

src/View/Helper/FormHelper.php
```diff
@@ -191,6 +191,14 @@ public function input($fieldName, array $options = [])
...
+            $templater->push();
+            $templateMethod = is_string($options['templates']) ? 'load' : 'add';
+            $templater->{$templateMethod}($options['templates']);
+        }
+        unset($options['templates']);
```
jrble819: @dereuromark Did you mean this unset() at line 200?

dereuromark: Yes, !empty() and `[]` or any non-nullish value might end up being outputted in the input one maybe.

jrble819:
```php
public function input($fieldName, array $options = [])
    {
        $options += [
            'prepend' => null,
            'append' => null,
            'type' => null,
            'label' => null,
            'error' => null,
            'required' => null,
            'options' => null,
            'help' => null,
            'templates' => [],
            'templateVars' => []
        ];
        $options = $this->_parseOptions($fieldName, $options);
        $reset = $this->templates();

        $templater = $this->templater();
        if (!empty($options['templates'])) {
            $templater->push();
            $templateMethod = is_string($options['templates']) ? 'load' : 'add';
            $templater->{$templateMethod}($options['templates']);
        }
        unset($options['templates']);
```
The function always throws in an empty templates array even when the passed $options doesn't have it.  I was checking to see if it is !empty() because it's always going to be set.  This chunk of code comes from Cake's original Form Helper.  

The only thing I'm unsure of now is whether or not the ```unset($options['templates']);``` should be there or not.  Not sure if there would be a need to have those templates passed on through to Cake's Form Helper.

ADmad:
I don't understand the issue. How can you not use custom templates already when the options are passed to parent's input() call?

darensipes:
@ADmad The problem is in the switch case for radio it is grabbing the ```radioInlineFormGroup``` or ```radioNestingLabel``` before the options merge/overrides happen. Therefore, if you need to make a tiny tweak to the templates for radios you can not, because bootstrap-ui's FormHelper hasn't merged the template options as you said that happens in the parent.

I have `App/View/Helper/FormHelper` that extends `BootstrapUI\View\Helper\FormHelper`, if I wanted a function like below in my helper to work correctly or input from a template, the proposed changes from @jrble819 would solve this issue.

```php
public function radioButtonGroup($fieldName, $options = [])
   {
       if (!isset($options['label'])) {
           $options['label'] = '';
       }
       $options['type'] = 'radio';
       $options['templates'] = [
           'formGroup' => '{{input}}',
           'inputContainer' => '<div class="radio-button-group"><label class="control-label input-group">' . $options['label'] . '</label><div class="btn-group {{type}}{{required}}" data-toggle="buttons">{{content}}{{help}}</div></div>',
           'radioNestingLabel' => '{{hidden}}<label class="btn btn-default">{{input}}{{text}}</label>',
       ];
       return parent::input($fieldName, $options);
   }
```

This would generate bootstrap radios like found on: http://www.bootply.com/bTURSgzRua

ADmad:
@darensipes I think the following patch would fix the issue with radio templates

```diff
diff --git a/src/View/Helper/FormHelper.php b/src/View/Helper/FormHelper.php
index 7525019..7046db5 100644
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -205,14 +205,12 @@ class FormHelper extends Helper
                 $isInline = (isset($options['inline']) && $options['inline'] === true);
 
                 $templates = [];
-                if ($isInline && $this->_align !== 'horizontal') {
-                    $templates['formGroup'] = $this->templater()->get('radioInlineFormGroup');
+                if (empty($options['templates']['formGroup']) && $isInline && $this->_align !== 'horizontal') {
+                    $options['templates']['formGroup'] = $this->templater()->get('radioInlineFormGroup');
                 }
-                if (!$isInline) {
-                    $templates['nestingLabel'] = $this->templater()->get('radioNestingLabel');
+                if (empty($options['templates']['radioNestingLabel']) && !$isInline) {
+                    $options['templates']['radioNestingLabel'] = $this->templater()->get('radioNestingLabel');
                 }
-
-                $this->templater()->add($templates);
                 break;
             case 'select':
                 if (isset($options['multiple']) && $options['multiple'] === 'checkbox') {
```

Try it out and let me know. I don't think an extra level of template stacking is necessary.

jadb:
Continuing discussion on proposed solution in #118

PR #115, opened 2016-01-22
Clarified where bootstrap.php is located

jonlachmann:
This had me searching for a while because I was unsure about where this file was, I first thought it was a part of bootstrap-ui and not cakephp.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Using [Composer][composer]:
 composer require friendsofcake/bootstrap-ui:~0.3
 ```
 
-You then need to load the plugin. In `boostrap.php`, something like:
+You then need to load the plugin. In `bootstrap.php` (note that this file is a part of cakephp and located in the config folder), something like:
 
 ```php
 \Cake\Core\Plugin::load('BootstrapUI');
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,7 +30,7 @@ Using [Composer][composer]:
 composer require friendsofcake/bootstrap-ui:~0.3
 ```
 
-You then need to load the plugin. In `boostrap.php`, something like:
+You then need to load the plugin. In `bootstrap.php` (note that this file is a part of cakephp and located in the config folder), something like:
```
josegonzalez: `config/bootstrap.php` is probably better.

README.md
```diff
@@ -30,7 +30,7 @@ Using [Composer][composer]:
 composer require friendsofcake/bootstrap-ui:~0.3
 ```
 
-You then need to load the plugin. In `boostrap.php`, something like:
+You then need to load the plugin. In `bootstrap.php` (note that this file is a part of cakephp and located at config/bootstrap.php), something like:
```
ADmad: This file is part of your app not cakephp

PR #118, opened 2016-01-24
Fix inability to override templates through FormHelper::input() options.

ADmad:
Closes #113

@jrble819 @darensipes Please try out this patch (use "dev-issue-113" through composer) and confirm it resolves your issue.


jadb:
@jrble819 @darensipes have you tried it? does it work?

@ADmad can't we have tests for it?

darensipes:
Hey been out on paternity leave. I will give it a check and make sure it works and let you know the results.

darensipes:
This works great it solves the issue from #113.

PR #123, opened 2016-02-08
CDN'ed

balcsida:
Replace local stylesheet and script files with CDN ones so those, who wants to use the original Bootstrap styles (without any modification) they could use it without any download/bower "magic".


ADmad:
I prefer not having the maintenance overhead of updating these links each time there is a new release of bootstrap / jQuery.

dereuromark:
Yeah I think if people would like to use them, they can have their own template here.
Usually you use bower and manage the versions yourself in your repo.

balcsida:
My main purpose was to make it more automated so people who don't want to use bower (not me!) could kick in a project a little faster. But @ADmad opinion seems fair to me, so I close this PR and will try to make a composer install script instead.

PR #124, opened 2016-02-08
Fixing the reset of the templates

darensipes:
Need to reset the templates otherwise when calling successive inputs with ```prepend``` option you get multiple appended HTML on subsequent inputs. It could possibly be affecting other options.

Example:
```php
<?= $this->Form->input('username', ['label' => 'Username or Email', 'prepend' => '<i class="fa fa-at"></i>', 'autocomplete' => 'off']) ?>
<?= $this->Form->input('password', ['prepend' => '<i class="fa fa-key"></i>', 'autocomplete' => 'off']) ?>
```

screen shot 2016-02-08 at 12 15 05 pm

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -189,6 +189,7 @@ class FormHelper extends Helper
             'templateVars' => []
         ];
         $options = $this->_parseOptions($fieldName, $options);
+        $reset = $this->templates();
 
         $newTemplates = $options['templates'];
         if ($newTemplates) {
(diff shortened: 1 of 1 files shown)


ADmad:
Hmm this is cause by BasicWidget modifying the templates. Ideally widgets shouldn't modify templates or at least revert to previous state.

darensipes:
This problem was caused by Issue #113 & #118. Not sure I follow with respect to the widgets.

ADmad:
The real problem is [here](https://github.com/FriendsOfCake/bootstrap-ui/blob/master/src/View/Widget/BasicWidget.php#L56). Widgets should not modify templates.

ADmad:
Another problem [here](https://github.com/FriendsOfCake/bootstrap-ui/blob/master/src/View/Widget/CheckboxWidget.php#L33). CheckboxWidget should definitely not modify `checkboxContainer` template which itself doesn't even use.

The backup and restoring of templates done earlier in FormHelper::input() was hiding these issues with widgets.

ADmad:
I will have a fix soon.

PR #125, opened 2016-02-09
Reset templates when extra options are used.

ADmad:
I have done a quick fix for now. Later will do a proper fix and prevent widgets from mucking up the templates.

Closes #124.


ADmad:
@darensipes Please try out this patch and let me if it solves your issue.

darensipes:
Yes, this solves the issue. Thanks!

PR #128, opened 2016-02-13
Use codecov.

ADmad:


composer.json
```diff
@@ -20,8 +20,7 @@
         "cakephp/cakephp": "~3.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "4.1.*",
-        "cakephp/cakephp-codesniffer": "dev-master"
```
josegonzalez: Why not leave this in here?

ADmad: To avoid multiple per project copies of codesniffer locally. Locally most have codesniffer installed globally and for travis it's installed for the phpcs build. We follow this in almost all core and foc repos.

jadb: They are part of the travis setup, and usually installed globally on dev machines. No?

josegonzalez: ^ what @jadb says. It's installed locally via composer cache, so I don't see the big deal with having it here. Also, I *don't* install anything globally if I can manage it - down that way leads failure...

Issue #129, opened 2016-02-18
Remove the default layout set from the trait


icaroscherma:
I suggest to remove the lines:
```PHP
if (empty($this->layout) || $this->layout === 'default') {
    $this->layout = 'BootstrapUI.default';
}
```
from `initializeUI()` in the `UIViewTrait`, because doing so you kind of force the user to create another template name and set it into its own `AppView`.
Or change the method to support some args and then enable, like:
```PHP
public function initializeUI(array $args)
{
    if ($args['useBsLayout'] && (empty($this->layout) || $this->layout === 'default')) {
        $this->layout = 'BootstrapUI.default';
    }
}
```
and inside your AppView you pass what you need: `$this->initializeUI(['useBsLayout' => true]);`.
=)

ADmad:
> because doing so you kind of force the user to create another template name and set it into its own AppView

I don't really understand what you mean. If you set a custom layout using viewBuilder() in your controller that will be used else by default 'BootstrapUI.default' will be used.

icaroscherma:
@ADmad in other terms, if you use the trait, you force the user to change the default layout's filename, otherwise it will use bootstrap's. ok, we both said the same thing, but I'm trying to say that this is not a good approach.

In my `default.ctp` layout, I've already put bootstrap's js/css and jquery, I just want to use the helpers from bootstrap-ui's plugin, by using the UIViewTrait the code become cleaner, because we just put one line describing that we want to enable this "resource" (the helpers), but it force me to change the `public $layout` to something else than "default", otherwise `vendor/.../default.ctp` will be used.

The idea behind the bootstrap-ui plugin is the code reuse (DRY), so the user keep using the helper but with bootstrap guidelines already inside it, not force the user to change its behavior, like default filenames.
(I know, this is too simple to change the "default.ctp" to another)

ADmad:
Okay, I understand now :) 

Would you like to make a PR which you proposed change? Should actually also remove the `empty($this->layout)` check and it would prevent rendering a template without layout.

PR #130, opened 2016-02-18
Fix issue #129

ayman-alkom:
Fix issue #129 
I will rebase it tommorow

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Transparently use [Twitter Bootstrap 3][twbs3] with [CakePHP 3][cakephp].
 Using [Composer][composer]:
 
 ```
-composer require friendsofcake/bootstrap-ui:~0.3
+composer require friendsofcake/bootstrap-ui
 ```
 
 Then load the plugin by adding the following to your app's `config/boostrap.php`:


src/View/UIViewTrait.php
```diff
@@ -13,11 +13,16 @@
...
+     * - `layout` to pass your layout, or it will use the plugin one
+     *
+     * @param array $options
      * @return void
      */
-    public function initializeUI()
+    public function initializeUI(array $options = [])
     {
-        if (empty($this->layout) || $this->layout === 'default') {
+        if (empty($options['layout'])) {
```
jadb: I think better to allow one to pass `layout = false` or `layout = ...`  if they don't want the default one or if they want a custom one instead. If `TRUE` is passed or option is not defined, use plugin's default.

Right now, `FALSE` will still define the plugin's layout and `TRUE` will just do nothing.

ayman-alkom: yes @jadb I thought like you, but I didn't found it meaningful. take another look, I may not right

jadb: Tried using it, I think the suggestion does make sense.

ayman-alkom:
Sorry @jadb the PR has been closed while i was trying to rebase it

PR #131, opened 2016-02-22
 UITrait: make it possible to pass a custom layout

### LINKED ISSUE #129: Remove the default layout set from the trait

ayman-alkom:
Fix #129

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ or using CakePHP's console:
 
 ## AppView Setup
 Then for a complete setup, just make your `AppView` class extends `BootstrapUI\View\UIView`.
-The `src\View\AppView.php` then will be as the following: 
+The `src\View\AppView.php` then will be as the following:
 
 ```php
 namespace App\View;
@@ -75,7 +75,7 @@ class AppView extends View
 {
 
     use UIViewTrait;
-    
+
     /**
      * Initialization hook method.
      */
--- a/src/View/UIView.php
+++ b/src/View/UIView.php
@@ -5,10 +5,10 @@ namespace BootstrapUI\View;
 use Cake\View\View;
 
 /**
- * UIView: the customised BootstrapUI View class.
+ * UIView: BootstrapUI customised view class.
  *
- * It customises the View::$layout to the BootstrapUI's layout and loads
- * BootstrapUI's helpers.
+ * - Customises the $layout to the BootstrapUI's one
+ * - Initialize the View with the BootstrapUI's helpers
  */
 class UIView extends View
 {
--- a/src/View/UIViewTrait.php
+++ b/src/View/UIViewTrait.php
@@ -3,20 +3,32 @@
 namespace BootstrapUI\View;
 
 /**
- * UIViewTrait: Trait that loads the custom UIBootstrap helpers and sets View
- * layout to the UIBootstrap's one.
+ * ## UIViewTrait:
+ *
+ * - loading the custom UIBootstrap helpers.
+ * - setting the View layout to the UIBootstrap's one.
  */
 trait UIViewTrait
 {
     /**
      * Initialization hook method.
      *
-     * @return void
+     * ### Options: an array of keys, available keys are:
+     * - `layout`:
+     *      - If empty or true will use the plugin's layout
+     *      - If a layout name passed it will be used
+     *      - If false do nothing ( will keep your layout )
+     *
+     * @return  void
      */
-    public function initializeUI()
+    public function initializeUI(array $options = [])
     {
-        if (empty($this->layout) || $this->layout === 'default') {
+        if (!array_key_exists('layout', $options)) {
             $this->layout = 'BootstrapUI.default';
+        } elseif ($options['layout'] === true) {
+            $this->layout = 'BootstrapUI.default';
+        } elseif (is_string($options['layout'])) {
+            $this->layout = $options['layout'];
         }
 
         $this->loadHelper('Html', ['className' => 'BootstrapUI.Html']);
--- a/tests/TestCase/View/UIViewTraitTest.php
+++ b/tests/TestCase/View/UIViewTraitTest.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace BootstrapUI\View;
+
+use Cake\TestSuite\TestCase;
+
+class UIViewTraitTest extends TestCase
+{
+    use UIViewTrait;
+
+    /**
+     * setUp method
+     *
+     * @return void
+     */
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->View = new UIView();
+    }
+
+    /**
+     * tearDown method
+     *
+     * @return void
+     */
+    public function tearDown()
+    {
+        parent::tearDown();
+        unset($this->View);
+    }
+
+    /**
+     * testInitializeUI method
+     *
+     * @return void
+     */
+    public function testInitializeUI()
+    {
+        $this->View->layout = 'default';
+        $this->View->initializeUI();
+        $this->assertEquals('BootstrapUI.default', $this->View->layout);
+
+        $this->View->layout = 'default';
+        $this->View->initializeUI([
+            'layout' => true
+        ]);
+        $this->assertEquals('BootstrapUI.default', $this->View->layout);
+
+        $this->View->layout = 'default';
+        $this->View->initializeUI([
+            'layout' => 'myLayout'
+        ]);
+        $this->assertEquals('myLayout', $this->View->layout);
+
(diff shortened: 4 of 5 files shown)


src/View/UIView.php
```diff
@@ -17,12 +17,14 @@ class UIView extends View
     /**
      * Initialization hook method.
      *
+     * @param array $options
      * @return void
      */
-    public function initialize()
+    public function initialize(array $options = [])
```
ADmad: `View::initialize()` does not take any params, so this proposed change will generate errors.

ayman-alkom: yes that's right

ayman-alkom: then we can just use the trait for the custom layout, and the class will always use the plugin's one

src/View/UIView.php
```diff
@@ -17,12 +17,14 @@ class UIView extends View
     /**
      * Initialization hook method.
      *
+     * @param array $options
```
ADmad: This needs to be removed

src/View/UIView.php
```diff
@@ -17,12 +17,14 @@ class UIView extends View
...
      * @return void
      */
     public function initialize()
     {
         parent::initialize();
 
+        //render the initializeUI method from the UIViewTrait
```
ADmad: Please don't add inline comments

src/View/UIViewTrait.php
```diff
@@ -3,22 +3,35 @@
...
+     *
+     * @param array $options
      * @return void
      */
-    public function initializeUI()
+    public function initializeUI(array $options = [])
     {
-        if (empty($this->layout) || $this->layout === 'default') {
-            $this->layout = 'BootstrapUI.default';
+        if ($options['layout'] !== false) {
```
ADmad: This will generate error if `layout` key is not set.

src/View/UIViewTrait.php
```diff
@@ -3,22 +3,35 @@
...
+            if (is_string($options['layout']))
+                $this->layout = $options['layout'];
+            else
+                $this->layout = 'BootstrapUI.default';
         }
-
+        //Load BootstrapUI's helpers
```
ADmad: Remove inline comment

ADmad:
Fix the various coding standard errors as reported [here](https://travis-ci.org/FriendsOfCake/bootstrap-ui/jobs/110970894#L312)

ayman-alkom:
Hi @ADmad , what I have to do for codecov issue?

ADmad:
Add tests for UIView and UIViewTrait. Currently then don't have any coverage.

https://codecov.io/github/FriendsOfCake/bootstrap-ui/src/View/UIView.php?ref=5b9d15427ca51bd5277907e4cce92797c9350a06
https://codecov.io/github/FriendsOfCake/bootstrap-ui/src/View/UIViewTrait.php?ref=5b9d15427ca51bd5277907e4cce92797c9350a06

README.md
```diff
@@ -60,6 +60,7 @@ class AppView extends UIView
     public function initialize()
     {
         //Don't forget to call the parent::initialize()
+        //parent::initialize(['layout' => 'your_custom_layout']);
```
ADmad: This is not needed.

ayman-alkom:
ok :(

ayman-alkom:
Hi @ADmad it has been finished finally

src/View/UIViewTrait.php
```diff
@@ -3,20 +3,37 @@
...
-        if (empty($this->layout) || $this->layout === 'default') {
+        if (!array_key_exists('layout', $options)) {
+            $this->layout = 'BootstrapUI.default';
+        } elseif ($options['layout'] === true) {
             $this->layout = 'BootstrapUI.default';
+        } elseif (is_string($options['layout'])) {
+            $this->layout = $options['layout'];
+        }
+
+        $layout = (is_string(@$options['layout'])) ? $options['layout'] : 'BootstrapUI.default';
```
ADmad: I think line 34-37 are not needed and should be removed. All cases are already covered above.

tests/TestCase/View/UIViewTraitTest.php
```diff
@@ -0,0 +1,61 @@
...
+    public function testInitializeUI()
+    {
+        $this->View->initializeUI();
+        $this->assertEquals('BootstrapUI.default', $this->View->layout);
+
+        $this->View->initializeUI([
+            'layout' => true
+        ]);
+        $this->assertEquals('BootstrapUI.default', $this->View->layout);
+//
```
ADmad: Needs to be removed.

ayman-alkom:
It took a lot to be performed, because this is the first test I made, it is more simple than I thought

ayman-alkom:
Hi @ADmad I removed them

src/View/UIViewTrait.php
```diff
@@ -3,20 +3,29 @@
...
+     *      - If false do nothing ( will keep your layout )
+     *
+     * @return  void
      */
-    public function initializeUI()
+    public function initializeUI(array $options = [])
     {
-        if (empty($this->layout) || $this->layout === 'default') {
-            $this->layout = 'BootstrapUI.default';
+        $layout = (is_string(@$options['layout'])) ? $options['layout'] : 'BootstrapUI.default';
```
ADmad: You did the opposite of what I told you :(

ADmad: Use error silencing operator `@` is bad coding.

ADmad: Replace all your `$this->layout` assignment code with this:

```
        if (!isset($options['layout']) || $options['layout'] === true) {
            $this->layout = 'BootstrapUI.default';
        } elseif (!empty($options['layout'])) {
            $this->layout = $options['layout'];
        }
```

tests/TestCase/View/UIViewTraitTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->View->initializeUI([
+            'layout' => 'myLayout'
+        ]);
+        $this->assertEquals('myLayout', $this->View->layout);
+
+        $this->View->layout = 'other_layout';
+        $this->View->initializeUI([
+            'layout' => false
+        ]);
+        $this->assertEquals('other_layout', $this->View->layout);
```
ADmad: Also add a test case for `'layout' => ''`.

src/View/UIViewTrait.php
ayman-alkom: Yes I know ;)
You are right about the bad of using `@`, but I thought that more clean, I will retrieve them.

tests/TestCase/View/UIViewTraitTest.php
ayman-alkom: ok

ayman-alkom:
@ADmad  take a look

src/View/UIViewTrait.php
```diff
@@ -3,20 +3,32 @@
...
+     *      - If a layout name passed it will be used
+     *      - If false do nothing ( will keep your layout )
+     *
+     * @return  void
      */
-    public function initializeUI()
+    public function initializeUI(array $options = [])
     {
-        if (empty($this->layout) || $this->layout === 'default') {
+        if (!array_key_exists('layout', $options)) {
```
ADmad: You are really testing my patience. I gave the exact code to use and yet you did something different.

tests/TestCase/View/UIViewTraitTest.php
```diff
@@ -0,0 +1,69 @@
...
+
+        $this->View->layout = 'default';
+        $this->View->initializeUI([
+            'layout' => ''
+        ]);
+        $this->assertEquals('BootstrapUI.default', $this->View->layout);
```
ADmad: If layout option is empty string then view layout should also be empty string as its a valid value.

src/View/UIViewTrait.php
ayman-alkom: sorry for that @ADmad , you may do it by yourself as you want

Issue #133, opened 2016-03-09
Textarea not implemented


nicls:
Hello, 

it seems to me that the Form-Element 'textarea' is not implemented!?

Regards,
cls

ADmad:
What did you try?  What did you get? What did you expect?

thinkingmedia:
@nicls Don't use `$this->Form->textarea(...)` instead use `$this->Form->input('name',['type'=>'textarea']);`

This is the same for select options. It would be nice if the helper worked for specific input types.

PR #136, opened 2016-03-13
Add "form-control" class to inputs when using input specific methods.

ADmad:
Refs #133


ADmad:
Hmm this PR doesn't cover `select` tags generated by `time()`/`date()`/`datetime()` methods. I am wondering that instead of overriding these `FormHelper` methods it would be better to inject the "form-control" class in the `SelectWidget` instead.

PR #138, opened 2016-03-14
Add "form-control" class through widgets.

ADmad:
This allows the class to be added even when using input specific methods.
Refs #133.


lorenzo:
Good job!

PR #140, opened 2016-03-21
support for multiple classes in a string

CodeBlastr:
I gotta believe most people write their classes as a string instead of an array.  This little change just adds a check to see if someone put in an alternative way to name classes.  eg.  ["params" => ["class" => "alert alert danger alert-dismissible"]], where as now (without this change you have to put each class as a separate array element. eg.  ["params" => ["class" => ["alert", "alert-danger", "alert-dismissible"]]]


CodeBlastr:
Forget this, I found an issue with it which causes a php warning.

PR #141, opened 2016-03-21
out of the box alerts

CodeBlastr:
Not sure if it's just because I'm working with the cakedc/users plugin, but auth alerts were not working out of the box.  This update makes auth alerts work out of the box and/or play nice with the cakedc/users plugin with no downside that I can think of.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ composer require friendsofcake/bootstrap-ui
 Then load the plugin by adding the following to your app's `config/boostrap.php`:
 
 ```php
-\Cake\Core\Plugin::load('BootstrapUI');
+\Cake\Core\Plugin::load('BootstrapUI', ['bootstrap' => true]);
 ```
 
 or using CakePHP's console:
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -1,0 +1,5 @@
+<?php
+
+use Cake\Core\Configure;
+
+Configure::write('Auth.flash.params.class', ['alert', 'alert-danger', 'alert-dismissible']);


ADmad:
The use of config var `'Auth.flash.params.class'` is not part of core and specific only the CakeDC/Users plugins. Code specific to other plugins shouldn't be added to this plugin. I am reverting this merge. You can add the config to your app's bootstrap instead.

josegonzalez:
@ADmad Please don't. I merged as - at least to me - making plugins interact with each other nicely is a great way to make the development experience more seamless. That config var seemed pretty harmless overall, so adding it seemed like a good idea.

If anything, we could scope it to check if the other plugin is loaded, but thats about it.

ADmad:
The revert is done. Making inter plugin interaction easier is fine but not by adding no-op codes for all others, especially when it achieve without adding a one liner to your app instead of the plugin.

CodeBlastr:
It might be a one liner now, but it was a one liner that took me 3-5 hours to figure out how to do.

CodeBlastr:
I opened this issue https://github.com/CakeDC/users/issues/326 with the users plugin, but if you have some suggestions about what could be done differently so that one plugin is not effected by the other plugin (eg. the users plugin should respect theme settings)

ADmad:
The issue of auth error messages not being properly themed out of the box isn't specific to CakeDC/users plugin. The problem is Auth component sets ['class' => 'error'](https://github.com/cakephp/cakephp/blob/master/src/Controller/Component/AuthComponent.php#L446) when setting flash messages which overwrites the default classes `['alert', 'alert-dismissible', 'fade', 'in']` of BUI's FlashHelper. 

So you need to change the defaults of flash config of AuthComponent. The documentation can be updated to do so in a generic way.

ADmad:
The fact about having to modify Auth's  `flash` config key is already [documented](https://github.com/FriendsOfCake/bootstrap-ui#configuration).

CodeBlastr:
Yes I saw that in the documentation, what I spent a lot of time on was trying to figure out how it could be more plug and play.  The point of a theme is that you don't go through your other code and add a bunch of class definitions.  (and by the way that documentation still isn't all that clear).  

But I did confirm that removing the line ``'params' => ['class' => 'error']`` from the AuthComponent did make the theme work better.  Would you suggest that this line be removed from Cakephp?

ADmad:
> Yes I saw that in the documentation, what I spent a lot of time on was trying to figure out how it could be more plug and play. The point of a theme is that you don't go through your other code and add a bunch of class definitions. 

Not everything can be made to work just by installing the plugin. Almost any plugin will require a bit of configuration.

> (and by the way that documentation still isn't all that clear).

The code example and accompanying statement tells you exactly what needs to be done. If you think that's inadequate, you are welcome to suggest improvement.

> Would you suggest that this line be removed from Cakephp?

No I wouldn't. We can't modify the core just to make things a tad easier for users of this plugin. Changing the core code would cause problems to others who rely on existing defaults provided by core.

CodeBlastr:
> Not everything can be made to work just by installing the plugin. Almost any plugin will require a bit of configuration.

Well you and I are just going to disagree on this then :)  Plugins already require too much configuration in my opinion, and should be made plug and play whenever possible. 

> We can't modify the core just to make things a tad easier for users of this plugin.

It's not just this plugin.  It's any theme plugin.

Anyhow I opened the issue on cakephp, I'll wait to see if anyone else agrees that cake should respect theme settings more easily. Have good night, not sure what time it is where you are, but it's past 1:00 am here :+1:

PR #145, opened 2016-03-28
added github templates copied from cakephp

thinkingmedia:
These templates will help with new issues and PR

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,7 +4,7 @@ This is a (multiple allowed):
 * [ ] feature-discussion (RFC)
 
 * BootstrapUI Version: EXACT RELEASE VERSION (THE VERSION IS IN YOUR COMPOSER.LOCK FILE).
-* Twitter Bootstrap: FRAMEWORK VERSION
+* Bootstrap Framework Version: FRAMEWORK VERSION
 * jQuery: LIBRARY VERSION
 * Platform and Target: YOUR WEB-SERVER, DATABASE AND OTHER RELEVANT INFO AND HOW THE REQUEST IS BEING MADE, HERE.
 
@@ -19,4 +19,6 @@ EXPLAIN WHAT IS ACTUALLY HAPPENING, HERE.
 
 Before you open an issue, please check if a similar issue already exists or has been closed before.
 
-P.S. Remember, an issue is not the place to ask questions.
+P.S. Remember, an issue is not the place to ask questions. You can use [Stack Overflow](http://stackoverflow.com/questions/tagged/cakephp)
+for that or join the #cakephp channel on irc.freenode.net, where we will be more
+than happy to help answer your questions.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Expected Behavior
+EXPLAIN WHAT IS TO BE EXPECTED, HERE.
+
+### Actual Behavior
+EXPLAIN WHAT IS ACTUALLY HAPPENING, HERE.
+
+Before you open an issue, please check if a similar issue already exists or has been closed before.
+
+P.S. Remember, an issue is not the place to ask questions.
```
ADmad: If we are gonna remind people not to use issues for question then also need to tell them where they can ask questions instead :)

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request.
+
+If it fixes a bug or resolves a feature request, be sure to link to that issue.
+
+The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.
```
dereuromark: Some would disagree with that one, as it does not apply to a lot of use cases where practical diffs of PRs show a lot more than thousand pages of theoretical RFC discussion :D

.github/ISSUE_TEMPLATE.md
thinkingmedia: The original was from cakephp and it had links to stackoverflow and icr but the cakephp channels. I don't know where to send people for this plugin?

.github/PULL_REQUEST_TEMPLATE.md
thinkingmedia: This template is kind of late after the fact. It's shown when a PR is being created. People have already done the work for the PR. I wouldn't want to tell them not to submit.

What it should be is a template for what the PR message should be. For that I don't have a good example.

.github/ISSUE_TEMPLATE.md
ADmad: The same places.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
+This is a (multiple allowed):
+* [x] bug
+* [ ] enhancement
+* [ ] feature-discussion (RFC)
+
+* BootstrapUI Version: EXACT RELEASE VERSION (THE VERSION IS IN YOUR COMPOSER.LOCK FILE).
+* Twitter Bootstrap: FRAMEWORK VERSION
```
jadb: Please drop `Twitter` usage.

.github/PULL_REQUEST_TEMPLATE.md
jadb: Ideas don't have to be long RFCs, rather opener discussions on a certain suggestion (if said suggestion would take much more time than opening/discussing it for a few days). Otherwise, I agree with @dereuromark, in all other cases, a quick PR is encouraged.

Am just not sure how to best word it in the template. I'll leave that to you @thinkingmedia :wink:

thinkingmedia:
okay done. How are these versions?

PR #146, opened 2016-03-28
small fixes for unit tests

thinkingmedia:
One of the tests were failing for me on Windows (stupid line feeds).

While looking at the unit tests. PhpStorm was flagging various things with warnings. So I added definitions for private properties and fix a couple of grammars.

I noticed that there aren't any units for the widgets. Can widgets be tested? How would that be done?

### DIFF
--- a/tests/TestCase/View/Helper/FlashHelperTest.php
+++ b/tests/TestCase/View/Helper/FlashHelperTest.php
@@ -3,7 +3,6 @@
 namespace BootstrapUI\Test\TestCase\View\Helper;
 
 use BootstrapUI\View\Helper\FlashHelper;
-use Cake\Controller\Controller;
 use Cake\Network\Request;
 use Cake\Network\Session;
 use Cake\TestSuite\TestCase;
@@ -15,6 +14,15 @@ use Cake\View\View;
  */
 class FlashHelperTest extends TestCase
 {
+    /**
+     * @var View
+     */
+    private $View;
+
+    /**
+     * @var FlashHelper
+     */
+    private $Flash;
 
     /**
      * setUp method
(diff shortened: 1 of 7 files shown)


thinkingmedia:
never mind, I found the widget tests from cakephp. Doesn't look hard. I'll see if I can add a few.

tests/TestCase/View/Helper/FlashHelperTest.php
```diff
@@ -15,6 +14,15 @@
  */
 class FlashHelperTest extends TestCase
 {
+    /**
+     * @var View
+     */
+    private $View;
```
ADmad: We don't use privates and protected would need the var name to be prefixed with underscore. So just change all to `public`.

tests/TestCase/View/Helper/PaginatorHelperTest.php
```diff
@@ -50,6 +59,7 @@ public function setUp()
         Router::connect('/:controller/:action/*');
         Router::connect('/:plugin/:controller/:action/*');
 
+        // @todo Set locale for unit test without using private property.
```
ADmad: Todos almost always get left forever :)  Either fix the issue or remove the todo.

tests/TestCase/View/Helper/FlashHelperTest.php
thinkingmedia: Question: Why don't you use privates?

thinkingmedia:
am I suppose to post a comment to say I did the changes or does GitHub somehow notify people that I made the changes? Just wondering how automated this PR stuff is.

ADmad:
GH doesn't notify when new commits are added to PR. So a comment is fine.

PR #147, opened 2016-03-28
added button tests and support for btn-default as a default style.

### LINKED ISSUE #143: FormHelper->button() should set btn-default
```
    echo $this->Form->button('Sign In');
```

Above outputs

```
    <button type="submit" class="btn">Sign In</button>
```

Should be outputting with `btn-default`.

```
    <button type="submit" class="btn btn-default">Sign In</button>
```

But it should be smart enough to know when not to include it.

```
    echo $this->Form->button('Sign In',['class'=>'primary']);
```

Should output with `btn-primary` and not `btn-default`.

```
    <button type="submit" class="btn btn-primary">Sign In</button>
```

Since this could impact previous versions. I wanted to verify this was okay before doing a PR.

thinkingmedia:
This PR improves buttons.

- closes #143 as button widget now assigns `btn-default` if it's missing.
- ButtonWidget now checks for both `primary` and `btn-primary` classname patterns in class attribute.
- Adds unit test coverage for ButtonWidget
- Fixes `btn-default` for `Form->submit()` which didn't use ButtonWidget.

I added a public static method to ButtonWidget which is called from FormHelper. So that there is only one function that handles class styles for buttons. I'm not sure if a helper should call a static on a widget. Seems dirty to me. Is this okay for now?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,9 @@
         "cakephp/cakephp": "~3.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "4.1.*"
+        "phpunit/phpunit": "4.1.*",
+        "squizlabs/php_codesniffer": "2.*",
+        "cakephp/cakephp-codesniffer": "dev-master"
     },
     "support": {
         "issues": "http://github.com/friendsofcake/bootstrap-ui/issues",
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,6 @@
     },
     "require-dev": {
         "phpunit/phpunit": "4.1.*",
-        "squizlabs/php_codesniffer": "2.*",
         "cakephp/cakephp-codesniffer": "dev-master"
     },
     "support": {
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -2,6 +2,7 @@
 
 namespace BootstrapUI\View\Helper;
 
+use BootstrapUI\View\Widget\ButtonWidget;
 use Cake\Core\Configure\Engine\PhpConfig;
 use Cake\Utility\Hash;
 use Cake\View\Helper\FormHelper as Helper;
@@ -160,8 +161,8 @@ class FormHelper extends Helper
             $class = array_unique(array_merge($class, $optionsClass));
             unset($options['class']);
         }
-
         $options = Hash::merge($options, ['class' => $class]);
+        $options['class'] = ButtonWidget::applyStyle($options['class']);
 
         return parent::submit($caption, $options);
     }
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -155,15 +155,7 @@ class FormHelper extends Helper
      */
     public function submit($caption = null, array $options = [])
     {
-        $class = ['btn'];
-        if (array_key_exists('class', $options)) {
-            $optionsClass = is_array($options['class']) ? $options['class'] : explode(' ', $options['class']);
-            $class = array_unique(array_merge($class, $optionsClass));
-            unset($options['class']);
-        }
-        $options = Hash::merge($options, ['class' => $class]);
-        $options['class'] = $this->applyStyle($options['class']);
-
+        $options = $this->applyButtonStyles($options);
         return parent::submit($caption, $options);
     }
 
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -154,7 +154,7 @@ class FormHelper extends Helper
      */
     public function submit($caption = null, array $options = [])
     {
-        $options = $this->applyButtonStyles($options);
+        $options = $this->applyButtonClasses($options);
         return parent::submit($caption, $options);
     }
 
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -162,7 +162,7 @@ class FormHelper extends Helper
             unset($options['class']);
         }
         $options = Hash::merge($options, ['class' => $class]);
-        $options['class'] = ButtonWidget::applyStyle($options['class']);
+        $options['class'] = $this->applyStyle($options['class']);
 
         return parent::submit($caption, $options);
     }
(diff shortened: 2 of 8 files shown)


tests/TestCase/View/Widget/ButtonWidgetTest.php
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * ButtonWidgetTest class
+ *
+ */
+class ButtonWidgetTest extends TestCase
+{
+    /**
+     * @var StringTemplate
+     */
+    private $templates;
```
ADmad: Change to `public`.

src/View/Helper/FormHelper.php
```diff
@@ -160,8 +161,8 @@ public function submit($caption = null, array $options = [])
             $class = array_unique(array_merge($class, $optionsClass));
             unset($options['class']);
         }
-
         $options = Hash::merge($options, ['class' => $class]);
+        $options['class'] = ButtonWidget::applyStyle($options['class']);
```
ADmad: Don't like this. You can add the method `applyStyle()` to `OptionsAwareTrait`. That trait is already used in the helpers and widgets.

thinkingmedia: OptionsAwareTrait is part of CakePHP.. That method sets the specific styles for bootstrap.

thinkingmedia: Maybe I could add a new trait to this project?

ADmad: @thinkingmedia I am referring to this one https://github.com/FriendsOfCake/bootstrap-ui/blob/master/src/View/Helper/OptionsAwareTrait.php :)

thinkingmedia: WTF! I must be getting old. I thought that was in cake. lol

ADmad: There's one with same classname in cake too but under different namespace for different purpose :)

thinkingmedia:
I had to change this in `composer.json` because I wanted to use the `pre-commit` hook for git.

```
    "require-dev": {
        "phpunit/phpunit": "4.1.*",
        "squizlabs/php_codesniffer": "2.*",
        "cakephp/cakephp-codesniffer": "dev-master"
    },
```

Now the hook is working.

thinkingmedia:
This PR has been update

- Adds more class functions to the OptionsAwareTrait
- Adds unit test for OptionsAwareTrait
- Button/Form updated to use OptionsAwareTrait for setting button styles.

composer.json
```diff
@@ -20,7 +20,9 @@
         "cakephp/cakephp": "~3.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "4.1.*"
+        "phpunit/phpunit": "4.1.*",
+        "squizlabs/php_codesniffer": "2.*",
```
ADmad: This is not needed. `cakephp/cakephp-condesniffer` already has it as a dependency.

composer.json
```diff
@@ -20,7 +20,9 @@
...
+        "cakephp/cakephp-codesniffer": "dev-master"
```
ADmad: Better to use `~2.0`.

src/View/Helper/OptionsAwareTrait.php
```diff
@@ -5,7 +5,96 @@
 trait OptionsAwareTrait
 {
     /**
-     * Injects classes into `$options['class']` when they don't already exist.
+     * A list of allowed styles for buttons.
+     *
+     * @var array
+     */
+    public $ButtonStyles = [
```
ADmad: `$buttonStyle` instead of `$ButtonStyles`.

ADmad: Also `$buttonClasses` would be better.

src/View/Helper/OptionsAwareTrait.php
```diff
@@ -5,7 +5,96 @@
...
+        'info', 'btn-info',
+        'primary', 'btn-primary'
+    ];
+
+    /**
+     * A mapping of aliases for button styles.
+     *
+     * @var array
+     */
+    public $ButtonStyleAliases = [
```
ADmad: `$ButtonStyleAliases` => `$buttonClassAliases`.

src/View/Helper/OptionsAwareTrait.php
```diff
@@ -5,7 +5,96 @@
...
+        'primary' => 'btn-primary'
+    ];
+
+    /**
+     * Contains the logic for applying style classes for buttons.
+     *
+     * @param array $data An array of HTML attributes and options.
+     * @return array An array of HTML attributes and options.
+     */
+    public function applyButtonStyles(array $data)
```
ADmad: `applyButtonClasses`

src/View/Widget/ButtonWidget.php
```diff
@@ -10,15 +10,6 @@ class ButtonWidget extends \Cake\View\Widget\ButtonWidget
 
     use OptionsAwareTrait;
 
-    protected $_styles = [
```
ADmad: Hmm.. Removing the property is not BC.

ADmad: Mark it deprecated instead.

thinkingmedia:
updated PR

- renamed public properties
- removed squizlabs/php_codesniffer
- update git hook script (copies from cakephp so it now does a php lint)

thinkingmedia:
updated PR

- restored property and made it deprecated

src/View/Helper/FormHelper.php
```diff
@@ -2,6 +2,7 @@
 
 namespace BootstrapUI\View\Helper;
 
+use BootstrapUI\View\Widget\ButtonWidget;
```
ADmad: This needs to be removed.

thinkingmedia:
updated PR

- removed BudgetWidget use from FormHelper

src/View/Helper/OptionsAwareTrait.php
ADmad: This method name change still needs to be addressed :)

thinkingmedia:
updated PR

- renamed method to applyButtonClasses

thinkingmedia:
boy that was suppose to be an easy change. lol

thinkingmedia:
bump

ADmad:
@jadb Thoughts?

PR #150, opened 2016-06-06
Allow the dashboard header to be set manually

glsherid:
## Description

Allow the dashboard header to be set manually instead of using the Controller name

Previously, the controller name was used for the dashboard page header. However if the controller was more than one word long then the header would be the Controller name without any spaces

### DIFF
--- a/src/Template/Layout/examples/dashboard.ctp
+++ b/src/Template/Layout/examples/dashboard.ctp
@@ -44,7 +44,9 @@ $this->start('tb_body_start');
                 <?= $this->fetch('tb_sidebar') ?>
             </div>
             <div class="col-sm-9 col-sm-offset-3 col-md-10 col-md-offset-2 main">
-                <h1 class="page-header"><?= $this->request->controller; ?></h1>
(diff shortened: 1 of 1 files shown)


dereuromark:
Wouldn't it make sense then to also Inflectlor::humanize() this default value?

glsherid:
@dereuromark humanise will replace underscored words, in my case the controller name appears in camel case. Do you know any methods to easily separate that?

dereuromark:
Humanize does separate via spaces :) See the test cases - or http://sandbox.dereuromark.de/sandbox/inflector?string=hello_world

glsherid:
$this->request->controller gives 'HelloWorld'. So it would take 2 lines - underscoring and then humanising. Is that what you mean?

dereuromark:
Was just an idea to resolve the multi word issue as a default value :)

glsherid:
Yes, it's a good idea! I'll make the change

dereuromark:
Any update?

ADmad:
Closing due to lack of follow up.

PR #156, opened 2016-09-09
Input group options for textarea

monsurhoq:
In this current stable version, input group facilities exist only for input but not for textarea. Although it not documented on bootstrap, but bootstrap also support input group addon for textarea. Please check http://stackoverflow.com/a/24558352/1787600

### DIFF
--- a/src/View/Widget/TextareaWidget.php
+++ b/src/View/Widget/TextareaWidget.php
@@ -31,8 +31,76 @@ class TextareaWidget extends \Cake\View\Widget\TextareaWidget
      */
     public function render(array $data, ContextInterface $context)
     {
-        $data = $this->injectClasses('form-control', $data);
+        $data += [
+            'type' => 'textarea',
+            'prepend' => null,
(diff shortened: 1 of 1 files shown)


ADmad:
Instead of all this code duplication can't what you want be achieved by just having TextAreaWidget extend `BootstrapUI\View\Widget\BaseWidget` instead of `\Cake\View\Widget\TextareaWidget`?

monsurhoq:
@ADmad, I also think so. But how can change data type `text` to `textarea` in `BootstrapUI\View\Widget\BasicWidget`? Data type replaced with `text` here.

ADmad:
May bad, since textarea generation is slightly different than other inputs just extending BasicWidget won't help. Still to avoid code duplication the common code can be moved in a trait which both BasicWidget and TextareaWidget can use or BasicWidget's `render()` can be split is to smaller protected method so that after TextaAreaWidget extends BasicWidget just `render()` can be overriden as required.

PR #159, opened 2016-09-22
Fix #148

### LINKED ISSUE #148: FormHelper select multiple with checkboxes produces duplicate checkboxes
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* BootstrapUI Version: 0.5.0
* Bootstrap Framework Version: 3.3.6
* jQuery: 2.1.4
* Platform and Target: CakePHP 3.2.8

### What you did
Following [CakePHP documentation](http://book.cakephp.org/3.0/en/views/helpers/form.html#creating-select-pickers), this was added in the edit.ctp.
```php
$options = [
    'Value 1' => 'Label 1',
    'Value 2' => 'Label 2'
];

echo $this->Form->select('field', $options, ['multiple' => 'checkbox']);
```
If BootstrapUI FormHelper is not loaded the checkboxes appear correctly.

### Expected Behavior
A list of checkboxes with labels

### Actual Behavior
A list of **double** checkboxes with labels

fireshot capture 29 - roles_ - http___localhost_8010_roles-capabi

dereuromark:
Resolves https://github.com/FriendsOfCake/bootstrap-ui/issues/148

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -31,7 +31,6 @@ class FormHelper extends Helper
         'help' => '<div class="help-block">{{content}}</div>',
         'inputContainer' => '<div class="form-group {{type}}{{required}}">{{content}}{{help}}</div>',
         'inputContainerError' => '<div class="form-group {{type}}{{required}} has-error">{{content}}{{error}}{{help}}</div>',
-        'checkboxWrapper' => '<div class="checkbox"><label>{{input}}{{label}}</label></div>',
         'multipleCheckboxWrapper' => '<div class="checkbox">{{label}}</div>',
         'radioInlineFormGroup' => '{{label}}<div class="radio-inline-wrapper">{{input}}</div>',
         'radioNestingLabel' => '<div class="radio">{{hidden}}<label{{attrs}}>{{input}}{{text}}</label></div>',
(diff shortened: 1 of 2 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -31,7 +31,6 @@ class FormHelper extends Helper
         'help' => '<div class="help-block">{{content}}</div>',
         'inputContainer' => '<div class="form-group {{type}}{{required}}">{{content}}{{help}}</div>',
         'inputContainerError' => '<div class="form-group {{type}}{{required}} has-error">{{content}}{{error}}{{help}}</div>',
-        'checkboxWrapper' => '<div class="checkbox"><label>{{input}}{{label}}</label></div>',
         'multipleCheckboxWrapper' => '<div class="checkbox">{{label}}</div>',
```
ADmad: I think `multipleCheckboxWrapper` can also be removed. It doesn't seem to be used by any widget in this plugin or core widgets. Only other reference I can find is [here](https://github.com/FriendsOfCake/bootstrap-ui/blob/master/src/View/Helper/FormHelper.php#L224) where it's overwritten by string contained in `checkboxWrapper`. So that particular line can also be removed.

dereuromark: That should better become a separate cleanup issue/task.
This fix here is isolated and well tested, whereas the multipleCheckboxWrapper thing seems to be somewhat entangled in the current code still:

```php
case 'select':
                    $options['templates']['checkboxWrapper'] = $this->templater()->get('multipleCheckboxWrapper');
                    $options['type'] = 'multicheckbox';
                }
                break;
```

PR #162, opened 2016-09-28
Input group options for textarea

monsurhoq:
Please check Input group options for textarea #156

### DIFF
--- a/src/View/Widget/InputgroupTrait.php
+++ b/src/View/Widget/InputgroupTrait.php
@@ -1,0 +1,102 @@
+<?php
+namespace BootstrapUI\View\Widget;
+
+use BootstrapUI\View\Helper\OptionsAwareTrait;
+use Cake\View\Form\ContextInterface;
+
+trait InputgroupTrait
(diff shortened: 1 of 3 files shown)


dereuromark:
This would need a test case.

src/View/Widget/InputgroupTrait.php
```diff
@@ -0,0 +1,102 @@
...
+     * Any other keys provided in $data will be converted into HTML attributes.
+     *
+     * @param array $data The data to build an input with.
+     * @param \Cake\View\Form\ContextInterface $context The current form context.
+     * @return string
+     */
+    public function output(array $data, ContextInterface $context)
+    {
+        $data += [
+            'type' => 'text',
```
ADmad: This will incorrectly add `'type'` attribute to the `textarea` tag. It should be removed and instead set only in the `BasicWidget` before this method is called.

dereuromark:
Follow up PR now open. Thanks a lot for your initial work here!

PR #164, opened 2016-10-18
A tiny change to README.md#Installing

dvlprs:
## Description

Made it obvious one has to `cd` into their CakePHP directory and run the composer command from it.
## Summarize

Edited `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,9 @@ Transparently use [Bootstrap 3][twbs3] with [CakePHP 3][cakephp].
 - Sample layouts (cover, signin, dashboard)
 - Bake templates *incomplete*
 
-## Installing
+## Installing Using [Composer][composer]
 
-Using [Composer][composer]:
+Go to your CakePHP directory and run the following command:
 
 ```
 composer require friendsofcake/bootstrap-ui
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Transparently use [Bootstrap 3][twbs3] with [CakePHP 3][cakephp].
 
 ## Installing Using [Composer][composer]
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,9 +23,9 @@ Transparently use [Bootstrap 3][twbs3] with [CakePHP 3][cakephp].
 - Sample layouts (cover, signin, dashboard)
 - Bake templates *incomplete*
 
-## Installing
+## Installing Using [Composer][composer]
 
-Using [Composer][composer]:
+`cd` to your CakePHP directory and run the following command:
```
ADmad: Yeah "your app" or "your project" directory would be more appropriate.

dereuromark:
I am not a too big fan of this, as it should be very clear when working with composer where to be for it to run (where the composer.json file is, so in the app dir).
This seems to be redundant and clear from the scope.

But meh, if it helps some people..

dvlprs:
> But meh, if it helps some people..

My thoughts exactly. There's a chance someone will stop for a second and be uncertain about that. Not everyone has as much experience with composer as you do for instance. I'm unsure what exactly is wrong with saying "*your CakePHP directory*", but if you feel it's not okay then so be it. Rephrased that portion.

dereuromark:
CakePHP dir = core dir (vendor/cakephp/cakephp) = dangerous tip.
My remark wasn't about rephrasing, I was just not sure this additional note is necessary.

dvlprs:
> CakePHP dir = core dir (vendor/cakephp/cakephp) = dangerous tip.

I see what you're saying. Sorry about that then.

PR #170, opened 2017-01-13
Pagination previous link not aligned

hiryu85:
Bootstrap3 need classname "previous" instead "prev"


dereuromark:
According to http://getbootstrap.com/components/#pagination this looks legit.
I wonder why in all those years no one else so far stumbled upon this issue.

PR #172, opened 2017-01-15
Input group

ADmad:
This allows using [input group](http://getbootstrap.com/components/#input-groups) with select and textarea tags too.

Refs #156, #162


dereuromark:
Not sure if removing OptionsAwareTrait (instead of keeping it around) could be BC breaking in case people further extended the helpers here.
But other than that 👍

ADmad:
The `OptionsAwareTrait` is now used by the new `InputGroupTrait` so there won't be any BC break.

PR #173, opened 2017-01-15
Check if the help option is set

cleptric:
Check if the help option is set to prevent `Undefined index` notices

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -351,7 +351,7 @@ class FormHelper extends Helper
             'label' => $options['label'],
             'error' => $options['error'],
             'templateVars' => isset($options['options']['templateVars']) ? $options['options']['templateVars'] : [],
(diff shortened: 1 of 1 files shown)


dereuromark:
In what case can it not be set?

cleptric:
@dereuromark I updated my app to cakephp 3.4-rc1 and used the `Form::control` method.

dereuromark:
@ADmad How can we test in travis against 3.next?
I think it would be in general a good idea to have plugins report early on possible bc breaks here - we can have them in the allowed failure list if necessary, but at least to know in advance what needs to be fixed.

cleptric:
screen shot 2017-01-15 at 23 52 51

dereuromark:
We should try to add the missing `control()` method and get that working.

dereuromark:
Closing as PR to fix the source of the issue is now open. Thanks for the quick report here.

PR #175, opened 2017-01-15
Fix new 3.4 control() to work properly.

dereuromark:
Fixes https://github.com/FriendsOfCake/bootstrap-ui/pull/173

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -171,7 +171,8 @@ class FormHelper extends Helper
      * @return string Completed form widget.
      * @deprecated Use control() instead.
      */
-    public function input($fieldName, array $options = []) {
+    public function input($fieldName, array $options = [])
+    {
         return $this->control($fieldName, $options);
     }
 
@@ -264,7 +265,12 @@ class FormHelper extends Helper
             );
         }
 
-        $result = parent::control($fieldName, $options);
+        if (method_exists('Cake\View\Helper\FormHelper', 'control')) {
+            $result = parent::control($fieldName, $options);
+        } else {
+            $result = parent::input($fieldName, $options);
+        }
+
         if ($newTemplates) {
             $this->templater()->pop();
         }
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -25,7 +25,7 @@ class FormHelper extends Helper
      *
      * @var string
      */
-    protected $_controlMethod;
+    protected $_controlMethod = 'input';
 
     /**
      * Default Bootstrap string templates.
@@ -109,7 +109,6 @@ class FormHelper extends Helper
 
         $this->_defaultWidgets = $this->_widgets + $this->_defaultWidgets;
 
-        $this->_controlMethod = 'input';
         if (method_exists('Cake\View\Helper\FormHelper', 'control')) {
             $this->_controlMethod = 'control';
         }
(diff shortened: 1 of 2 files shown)


dereuromark:
Well, it's not quite the proper BC fix here :) 3.4 works now, but 3.3 and before don't.

src/View/Helper/FormHelper.php
```diff
@@ -169,8 +169,26 @@ public function submit($caption = null, array $options = [])
      * @param string $fieldName This should be "Modelname.fieldname".
      * @param array $options Each type of input takes different options.
      * @return string Completed form widget.
+     * @deprecated Use control() instead.
      */
-    public function input($fieldName, array $options = [])
+    public function input($fieldName, array $options = []) {
+        return $this->control($fieldName, $options);
```
josegonzalez: PHPCS violation here.

src/View/Helper/FormHelper.php
```diff
@@ -246,7 +264,7 @@ public function input($fieldName, array $options = [])
             );
         }
 
-        $result = parent::input($fieldName, $options);
+        $result = parent::control($fieldName, $options);
```
josegonzalez: Could do a `method_exists` check to add support for 3.3 and below.

ADmad:
Tests failing.

Adding the new `control()` method and maintaining BC without code duplication is gonna be tricky.

dereuromark:
Good idea, @josegonzalez

ADmad:
Nice, tests passing now. I'll try it in one of my apps to check for BC and then merge.

src/View/Helper/FormHelper.php
```diff
@@ -102,6 +109,11 @@ public function __construct(View $View, array $config = [])
 
         $this->_defaultWidgets = $this->_widgets + $this->_defaultWidgets;
 
+        $this->_controlMethod = 'input';
```
dereuromark: You could also default the property to "input" - and only overwrite here if necessary.
That would get rid of this line.

dereuromark: But shouldn't it be the other way around: control() now by default, and if !method exists we fall back to the old input?

josegonzalez: I didn't want to add the extra negation. Ever so slightly faster :P

dereuromark: For me it's more about the logical "thats the status quo" and "thats the old bc fallback" kind of way. the time speed is irrelevant anyway, as you well know :)

dereuromark: But no big deal either way.

ADmad:
Since the public API has changed it's prudent to bump minor version in new release right?  I just released 0.6.0 yesterday :stuck_out_tongue:

dereuromark:
It was just a method being added. I think a bugfix version would also be reasonable.
Either way, you are still below 1.0 - a new minor also doesn't hurt :)

dereuromark:
My question still stands: How can we test Cake plugins beforehand towards the 3.next in Travis?
That would solve most of those things when the next minor comes out.

ADmad:
> How can we test Cake plugins beforehand towards the 3.next in Travis?

Dynamically switch the core version in `.travi.yml` similar to how it's done [here](https://github.com/cakephp/cakephp/blob/3.next/.travis.yml#L91) for the codesniffer version.

dereuromark:
Shouldn't we promote that for all plugins for Cake, FriendsOfCake and the rest of the ecosystem?

ADmad:
If someone wants to put in the effort sure.

PR #181, opened 2017-01-30
set array index to empty array if it doesn't exist

tgrowden:
## Description
This serves to prevent debug errors by passing an empty array if the `$options['input']` key does not exist.

## Summarize
This serves to prevent debug errors by passing an empty array if the `$options['input']` key does not exist.


ADmad:
Duplicate of  #180

PR #192, opened 2017-04-12
fixing clearfix for FormHelper::submit in form-horizontal

### LINKED ISSUE #161: Form Helper generates unexpected html
This is a (multiple allowed):
- [x] bug
- [ ] enhancement
- [x] feature-discussion (RFC)
- BootstrapUI Version: 0.5.1
- Bootstrap Framework Version: 3.3.7
- jQuery: 3.1.1
- CakePHP: 3.3
### What you did

I'm using a commercial bootstrap theme (inspinia) and try to create a basic horizontal form.

```
<?= $this->Form->create(${$tableAlias}, ['align' => [
        'sm' => [
            'left' => 2,
            'middle' => 10,
            'right' => 12
        ],
        'md' => [
            'left' => 2,
            'middle' => 10,
            'right' => 4
        ]
    ]]); ?>
    <?= $this->Form->input('username'); ?>
    <?= $this->Form->input('email'); ?>
    <?= $this->Form->input('password'); ?>
    <?= $this->Form->input('first_name'); ?>
    <?= $this->Form->input('last_name'); ?>
    <?= $this->Form->input('active', ['type' => 'checkbox']); ?>
    <?= $this->Form->submit('Save', ['class' => 'btn-primary']); ?>
 <?= $this->Form->end() ?>
```

the generated html of all inputs is as expected except of the submit button.
### Expected Behavior

This should be generated:

```
<div class="form-group">
    <div class="col-sm-offset-2 col-md-offset-2 col-sm-10 col-md-10">
        <input type="submit" class="btn btn-primary" value="Save">
    </div>
</div>
```
### Actual Behavior

The generated  html output for the submit button

```
<div class="col-sm-offset-2 col-md-offset-2 col-sm-10 col-md-10">
<input type="submit" class="btn btn-primary" value="Save">
</div>
```

The examples on the official bootstrap site also says that there should be a form-group wrapper.
http://getbootstrap.com/css/#forms-horizontal

The issue I'm having is that because there is no wrapper I have to put a clear fix at the and of the form because the grid wrapper is floating.

lorenzomaieru:
## Description
when using `Form->submit()`, the resulting html failed to be wrapped in between `<div class="form-group"></div>`. This resulted in a missing clear fix when using form-horizontal, breaking the form.

## Summarize
This Pull Request adds the missing `form-group` wrapper and fixes the issue.

## Related Issues

fixes #161

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -63,7 +63,7 @@ class FormHelper extends Helper
             'label' => '<label class="control-label %s"{{attrs}}>{{text}}</label>',
             'formGroup' => '{{label}}<div class="%s">{{input}}{{error}}{{help}}</div>',
             'checkboxFormGroup' => '<div class="%s"><div class="checkbox">{{label}}</div>{{error}}{{help}}</div>',
-            'submitContainer' => '<div class="%s">{{content}}</div>',
+            'submitContainer' => '<div class="form-group"><div class="%s">{{content}}</div></div>',
             'inputContainer' => '<div class="form-group {{type}}{{required}}">{{content}}</div>',
             'inputContainerError' => '<div class="form-group {{type}}{{required}} has-error">{{content}}</div>',
             'checkboxContainer' => '<div class="form-group{{required}}">{{content}}</div>',
--- a/tests/TestCase/View/Helper/FormHelperTest.php
+++ b/tests/TestCase/View/Helper/FormHelperTest.php
@@ -1173,6 +1173,24 @@ class FormHelperTest extends TestCase
         $this->assertHtml($expected, $result);
     }
 
+    public function testHorizontalFormSubmit()
+    {
+        $this->Form->create($this->article, ['align' => 'horizontal']);
+
+        $result = $this->Form->submit('Submit');
+        $expected = [
+            'div' => ['class' => 'form-group'],
+            ['div' => ['class' => 'col-md-offset-2 col-md-6']],
+            'input' => [
+                'type' => 'submit',
+                'value' => 'Submit',
+                'class' => 'btn btn-default',
+            ]
+        ];
+
+        $this->assertHtml($expected, $result);
+    }
+
     public function testStyledButton()
     {
         $result = $this->Form->button('Submit', ['class' => 'success']);


src/View/Helper/FormHelper.php
```diff
@@ -63,7 +63,7 @@ class FormHelper extends Helper
             'label' => '<label class="control-label %s"{{attrs}}>{{text}}</label>',
             'formGroup' => '{{label}}<div class="%s">{{input}}{{error}}{{help}}</div>',
             'checkboxFormGroup' => '<div class="%s"><div class="checkbox">{{label}}</div>{{error}}{{help}}</div>',
-            'submitContainer' => '<div class="%s">{{content}}</div>',
+            'submitContainer' => '<div class="form-group"><div class="%s">{{content}}</div></div>',
```
ADmad: I don't think 2 divs are necessary, you can just use `class="form-group %s"`

lorenzomaieru: Tried that locally, doesn't work.
The official documentation shows 2 divs: http://getbootstrap.com/css/#forms-horizontal

thinkingmedia:
This assumes that the developer wants the submit button rendered as a single row in the form. If you want to show a "submit" and "cancel" together then you can't with this chance.

This might fix your issue with a horizontal form, but it's going to break other people's forms.

dereuromark:
I agree with @thinkingmedia that it only serves one use case.
However, this would be the default one for 99% of the (baked) forms, thus this fix here seems to be a good default for now.

Ideally, we provide a single and a multi button template somehow, maybe with an additional method?
Then we could easily provide a way to have multiple buttons inside a form-group div.

thinkingmedia:
@dereuromark I read the original issue just now, and I can see how people expect this to just work out of the box.

Form layouts are problematic at best no matter what theme you use, and the bootstrap documentation do not show a two button form as an example. So I think this PR is fine.

If you want something different write the HTML yourself.

ADmad:
@thinkingmedia If people need flexibility they can use `Form->button()` which doesn't generate any wrapping div.

src/View/Helper/FormHelper.php
ADmad: Okay.

ADmad:
@lorenzomaieru Tests needed for the change.

ADmad:
@thinkingmedia Also this changes the template used only for horizontal form. `$_templateSet` maintains different sets of template which need to be changed depending on horizontal or inline form.

dakota:
https://github.com/FriendsOfCake/bootstrap-ui/pull/192#issuecomment-293897952

In that case this will need to be a new major version no? This will be quite a breaking change for people using the current behaviour?

dereuromark:
@dakota This could be part of a minor for the very recent 1.0 release. So far most are still using the pre releases I assume.

ADmad:
@dakota I don't think this change warrants a major release, a minor at best. The current default behavior is already broken for horizontal forms IMO.

lorenzomaieru:
> Tests needed for the change.
I wasn't sure whether to name the test `testHorizontalFormSubmit` or `testSubmitHorizontal`, as there are both `testHelpTextHorizontal` and `testBasicFormSubmit` in the rest of the file. 

Is there a naming convention for these?

dereuromark:
@lorenzomaieru I think that is fine, as long as they are readable.

Issue #176, opened 2017-01-18
CakePHP3 HtmlHelper::getCrumbList() is deprecated since 3.3.6


LiaraAlis:
Since CakePHP 3.3.6 the breadcrumbs helper should be used.
Maybe it is a good idea to add a BootstrapUI breadcrumbs helper based on the core breadcrumbs helper?

dereuromark:
Sounds great. A PR would be welcome.

PR #194, opened 2017-05-30
Adds support Breadcrumbs Helper

it-mob:
## Description
Adds support Breadcrumbs Helper

## Summarize
New Breadcrumb Helper
New BreadcrumbHelperTest

## Related Issues
#176


src/View/UIView.php
```diff
@@ -14,6 +14,7 @@
  * @property \BootstrapUI\View\Helper\FormHelper $Form
  * @property \BootstrapUI\View\Helper\HtmlHelper $Html
  * @property \BootstrapUI\View\Helper\PaginatorHelper $Paginator
+ * @property \BootstrapUI\View\Helper\BreadcrumbsHelperHelper $Breadcrumbs
```
dereuromark: BreadcrumbsHelperHelper => one helper too much

src/View/Helper/BreadcrumbsHelper.php
```diff
@@ -0,0 +1,45 @@
...
+    protected $_defaultAttributes = [
+        'class' => 'breadcrumb'
+    ];
+
+    /**
+     * @inheritdoc
+     */
+    public function render(array $attributes = [], array $separator = [])
+    {
+        $attributes = $attributes + $this->_defaultAttributes;
```
dereuromark: you can also use +=

src/View/Helper/BreadcrumbsHelper.php
```diff
@@ -0,0 +1,45 @@
+<?php
+/**
```
ADmad: Please get rid of this header.

src/View/Helper/BreadcrumbsHelper.php
```diff
@@ -0,0 +1,45 @@
...
+ * Created by PhpStorm.
+ * User: ngirardet
+ * Date: 30.05.2017
+ * Time: 12:03
+ */
+
+namespace BootstrapUI\View\Helper;
+
+class BreadcrumbsHelper extends \Cake\View\Helper\BreadcrumbsHelper
```
ADmad: Add "use" statement at top and just use class name.

tests/TestCase/View/Helper/BreadcrumbsHelperTest.php
```diff
@@ -0,0 +1,51 @@
...
+    public function testCrumbList()
+    {
+        $result = $this->Breadcrumbs
+            ->add('jadb')
+            ->add('admad')
+            ->add('joe')
+            ->render();
+
+        $expected = '<ol class="breadcrumb"><li><span>jadb</span></li><li><span>admad</span></li><li><span>joe</span></li></ol>';
+        $this->assertEquals($expected, $result);
```
ADmad: More tests needed for the various other placeholders used in templates.

src/View/Helper/BreadcrumbsHelper.php
dereuromark: use ... as CoreBreadcrumbsHelper IMO

josegonzalez:
I made a few of the desired changes but it still needs tests, and also it fails on php 5.4 since the BreadcrumbsHelper is new-ish.

src/View/UIViewTrait.php
```diff
@@ -33,5 +33,8 @@ public function initializeUI(array $options = [])
         $this->loadHelper('Form', ['className' => 'BootstrapUI.Form']);
         $this->loadHelper('Flash', ['className' => 'BootstrapUI.Flash']);
         $this->loadHelper('Paginator', ['className' => 'BootstrapUI.Paginator']);
+        if (class_exists('\Cake\View\Helper\BreadcrumbsHelper')) {
```
ADmad: Why the class exists chexk?

josegonzalez: We'd have to drop php 5.4 support if we didn't include this, as the BreadcrumbsHelper was only introduced in CakePHP 3.3.6.

ADmad: Ah right, forgot about that.

PR #195, opened 2017-06-12
wrong path

jeffreyroberts:
## Description
fixing wrong path

## Summarize
changed path from TwitterBootstrap to examples

## Benefits
it works now


jeffreyroberts:
this might actually not be the case, im tired, ill check in the morning if this is real or not

josegonzalez:
@jeffreyroberts _bump_

ADmad:
Closing since there was no follow up.

PR #201, opened 2017-09-27
Added the option to load, or not load the helpers.

fr3nch13:
## Description
Added the option to load, or not load the helpers.

## Summarize
Changes to the UIViewTrait

## Benefits
Allows the developer to make their own html helper and extend BootstrapUI's helper if needed.

### DIFF
--- a/src/View/UIViewTrait.php
+++ b/src/View/UIViewTrait.php
@@ -16,6 +16,9 @@ trait UIViewTrait
      *      - If not set or true will use the plugin's layout
      *      - If a layout name passed it will be used
      *      - If false do nothing (will keep your layout)
+     *  - `loadHelpers`
+     *      - If true, load the BootstrapUI helpers (default)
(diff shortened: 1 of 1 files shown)


ADmad:
Without the statements to load helpers all that remains is an `if.. else` block. So if you want just that add that directly to your `AppView::initialize()`.

fr3nch13:
Good point. Technically I don't need to use the UIView or the UIViewTrait. It seems they're there as a quick option to load BootstrapUI. 

My route is that I have my own plugin that loads and uses the BootstrapUI with it's own helpers that extend the BootstrapUI helpers and the layouts also extend the bootstrap ones.

Example: 
```
namespace PygonGit\PluginCore\View\Helper;

class FormHelper extends \BootstrapUI\View\Helper\FormHelper
{

    public function initialize(array $config)
    {
        parent::initialize($config);

    }
}
```

PR #203, opened 2017-10-22
escape value in staticControl

aarubui:
Escape value by default in staticControl


ADmad:
Thanks. The phpstan failure seems unrelated.

Issue #193, opened 2017-05-23
The install process is way too complicated


notoriousturtle:
This is a (multiple allowed):
* [ ] bug
* [ ] enhancement
* [x] feature-discussion (RFC)

* BootstrapUI Version: Latest

### Feature request, easier install process, and easier custom layout install.
Can we have an easy install script or something, such as this:
https://gist.github.com/notoriousturtle/ea0198288ce46580aea09f11b1bfd01b

It would be good if there was an easy way of quickly dropping a new layout into cake as well. So modifying the ctp files used during the generation process. The current problem I see here is, these files are all over the place, and can be very error prone. It also takes a fair bit of time, especially if you're new to the whole process. It would be good if I could drop a new layout in with relative ease. 

Thanks.

davidyell:
What is easy for you would be annoying for me, as I only use the helpers. So I'd have to remove all the automated stuff once it's finished.

Perhaps a shell task which does this would be an option? Perhaps optionally called from composer somehow?

davidyell:
I think the documentation could be improved the help people get up and running quicker.

ADmad:
The next major release with BS4 compatibility is going to have package.json to download all required libs from npm and cake shells to copy/symlink the assets into app's webroot.

davidyell:
So now the plugin depends on Node.js and NPM? That's far from ideal,
especially when Composer could do this.

dereuromark:
depends? no, it can be used with it, but you are not forced to.

davidyell:
That's great. I hope we can provide some install instructions which form a kind of Quick Start, if people just want to install the plugin, their own jQuery, and Bootstrap; then just get started. Without needing all the extras.

This was the issue raised in Slack. That the plugin lacked clear examples of implementation which didn't involve lots of complex extras. I thought I could sort that out, but it seems with such great progress on the Bootstrap 4 version, it's worth waiting for that release.

ADmad:
Installing jQuery and bootstrap's css and js on your own is no different than how you would use any another js/css in your app. Use whatever way you prefer to get your assets and put them in webroot and then link them in layout using helper.

davidyell:
Yes that would be ideal Quick Start docs content

ADmad:
Closing as v2 has a shell to install the assets (using npm).

PR #207, opened 2018-01-15
Feature/bootstrap4

### LINKED ISSUE #103: Bootstrap 4 Changes that cause this plug in to fail
We should focus on opening a new branch to prepare for the future:

Details can be seen here: http://v4-alpha.getbootstrap.com/migration/#forms

Bootstrap 4 failing with this plug in can be demoed here:
http://sandbox3.dereuromark.de/sandbox/bootstrap

In the Demo none of the bootstrap 4 js is included as of currently. The minified bootstrap css is however included and can be seen to not work with current plug in since bootstrap decided to change some class names.

julianpollmann:
## Description
Added Bootstrap4 support to this plugin.

## Summarize

- changed templates for ui components according to bootstrap4
- install bootstrap assets via npm and added post-install-cmd to composer.json to simplify the install process
- changed bake templates according to bootstrap4
- changed default and sample layouts
- removed components, which are deprecated in bootstrap4. See [Migrating to v4](https://getbootstrap.com/docs/4.0/migration/).

## Further Changes
- FormHelper's inline templates may be reworked.
- Since Bootstrap4 dropped support for [glyphicons](https://getbootstrap.com/docs/4.0/migration/#components), `HtmlHelper's icon()` function needs to be reworked.

## Related Issues

fixes #103, #193


ADmad:
This is a good start. You'll need to rebase due to changes on master.

It would be better to implement `post-install.sh` as a cake shell instead. Also I wouldn't run the script/shell as composer post install script. Someone might have their own setup to manage assets. Having a shell would allow the flexibility to use it only if required.

src/Template/Bake/Template/add.ctp
```diff
@@ -0,0 +1,76 @@
+<%
```
ADmad: Bake templates using erb style `<%` tags are deprecated. The new templates need to use twig syntax.

dereuromark: That will make migration a bit more difficult, also if stuff needs to be backported, fixes etc.
I wonder if we should continue using the deprecation in this case.
A rewrite in twig makes more sense for complete new projects or plugins and their bake templates IMO at this point.
The old style wont go anywhere soon - especially during 3.x.

src/View/Helper/FormHelper.php
```diff
@@ -59,24 +60,24 @@ class FormHelper extends Helper
...
             'submitContainer' => '<div class="form-group"><div class="%s">{{content}}</div></div>',
             'inputContainer' => '<div class="form-group {{type}}{{required}}">{{content}}</div>',
-            'inputContainerError' => '<div class="form-group {{type}}{{required}} has-error">{{content}}</div>',
+            'inputContainerError' => '<div class="form-group {{type}}{{required}}">{{content}}</div>',
             'checkboxContainer' => '<div class="form-group{{required}}">{{content}}</div>',
-            'checkboxContainerError' => '<div class="form-group{{required}} has-error">{{content}}</div>',
+            'checkboxContainerError' => '<div class="form-group{{required}}">{{content}}</div>',
             'radioContainer' => '<div class="form-group{{required}}">{{content}}</div>',
-            'radioContainerError' => '<div class="form-group{{required}} has-error">{{content}}</div>',
+            'radioContainerError' => '<div class="form-group{{required}}">{{content}}</div>',
```
dereuromark: Is the removal of has-error desired? Might take away some of the functionality of CSS styling in the error case?

julianpollmann: The Bootstrap Migration Guide mentions, that `.has-error` has been dropped in favor of `:invalid`/ `:valid` pseudo classes. See https://getbootstrap.com/docs/4.0/migration/#forms-1
For server side validation they suggest using `.is-invalid` and `.is-valid` on the wrapping element. See https://getbootstrap.com/docs/4.0/components/forms/#how-it-works

src/Template/Bake/Template/add.ctp
ADmad: I don't see any migration issue in relation to bake templates. When baking one doesn't really care what syntax the templates use. If someone has overridden the templates they are going to have to update them anyway for bootstrap 4 and in that case they can use which syntax they prefer.

src/Shell/Task/TwbsAssetsTask.php
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * Task for installing Bootstrap assets and layouts to app.
+ */
+class TwbsAssetsTask extends Shell
+{
+    private $_assetDir;
+    private $_nodeDir;
+    private $_cssDir;
+    private $_jsDir;
```
dereuromark: use protected. People might need to adjust those in an extending class.

src/Shell/Task/TwbsAssetsTask.php
```diff
@@ -0,0 +1,155 @@
...
+        $this->out('<success>Sample layouts copied successfully.</success>');
+    }
+
+    /**
+     * Copy files to assetdir's css/js path
+     *
+     * @param array $files Assetfiles to copy
+     * @return void
+     */
+    private function _copy(array $files)
```
dereuromark: same here

README.md
```diff
@@ -5,19 +5,21 @@
...
+* npm 5.x
+* jQuery 3.2+
+* Popper.js 1.x
 
 ## What's included?
 
-- FlashComponent + elements (types: error, info, success, warning)
-- FormHelper (align: default, inline, horizontal)
-- HtmlHelper (components: breadcrumbs, badge, label, icon)
+- FlashComponent + elements (types: primary, secondary, success, danger, warning, info, light, dark)
```
josegonzalez: Can we keep an error and info alias? And just mark as deprecated. Will ease the transition.

README.md
```diff
@@ -235,23 +228,27 @@ echo $this->Form->control('published', ['type' => 'checkbox']);
...
 
 ```html
-<form method="post" accept-charset="utf-8" role="form" class="form-horizontal" action="/articles/add">
-    <div style="display:none;"><input type="hidden" name="_method" value="POST"></div>
-    <div class="form-group">
-        <label class="control-label col-sm-6 col-md-4" for="title">Title</label>
-        <div class="col-sm-6 col-md-4">
-            <input type="text" name="title" required="required" id="title" class="form-control">
-        </div>
+<form method="post" accept-charset="utf-8" class="form-horizontal" role="form" action="/users/add">
```
josegonzalez: example still uses `$article`

README.md
dereuromark: How do the controller calls look like?
I think we should actually keep the 4 types we always had (error, info, success, warning), and instead allow the others to be opt-in on top if desired.
But error in cakephp is just internally transferred into a style class called "danger". But leaking those classes into the top layer is not good IMO.

ADmad: Yes, `error`, `info`, `success`, `warning` should be the the only valid types. Just the FlashHelper should set associated class name based on these 4 types.

README.md
```diff
@@ -31,17 +33,12 @@ Transparently use [Bootstrap 3][twbs3] with [CakePHP 3][cakephp].
 composer require friendsofcake/bootstrap-ui
 ```
 
-Then load the plugin by adding the following to your app's `config/boostrap.php`:
+Composers `post-install-cmd` will execute a script which will:
```
ADmad: This text change needs to be reverted / updated.

README.md
```diff
@@ -190,27 +180,30 @@ is from the BootstrapUI plugin.
 ### Basic Form
 
 ```php
-echo $this->Form->create($article);
+echo $this->Form->create($user);
```
ADmad: Should be reverted to `$article`.

README.md
```diff
@@ -190,27 +180,30 @@ is from the BootstrapUI plugin.
...
-        <input type="text" name="title" required="required" id="title" class="form-control">
-    </div>
-    <div class="form-group">
-        <input type="hidden" name="published" value="0">
-        <label for="published">
-            <input type="checkbox" name="published" value="1" id="published" class="form-control">
-            Published
-        </label>
-    </div>
+<form method="post" accept-charset="utf-8" role="form" action="/users/add">
```
ADmad: The URL should be reverted to `/articles/add`.

composer.json
```diff
@@ -20,7 +20,9 @@
         "cakephp/cakephp": "~3.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "<6.0"
+        "phpunit/phpunit": "<6.0",
+        "squizlabs/php_codesniffer": "^3.2",
```
ADmad: Explicitly adding `squizlabs/php_codesniffer` is not required since it's already a dependency of `cakephp/cakephp-codesniffer`.

src/Shell/Task/TwbsAssetsTask.php
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    /**
+     * Installs Bootstrap assets using npm
+     *
+     * @return void
+     */
+    public function installAssets()
+    {
+        $this->out('<info>Checking npm...</info>');
```
ADmad: You can just use `$this->info()` instead of `<info>` tag.

src/Shell/Task/TwbsAssetsTask.php
```diff
@@ -0,0 +1,155 @@
...
+        chdir(Plugin::path('BootstrapUI'));
+        $node_mod = new Folder('node_modules');
+        if ($node_mod->delete()) {
+            $this->out(__('<success>Cleared node_modules...</success>'));
+        }
+
+        exec('npm install --verbose', $output, $return);
+        $this->out($output);
+        if ($return === 0) {
+            $this->out('<success>Bootstrap assets installed successfully.</success>');
```
ADmad: Use `$this->success()`.

src/Shell/Task/TwbsAssetsTask.php
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Installs Bootstrap assets using npm
+     *
+     * @return void
+     */
+    public function installAssets()
+    {
+        $this->out('<info>Checking npm...</info>');
+        if (!`which npm`) {
+            $this->abort(__('NPM (https://www.npmjs.com/) is required, but not installed. Aborting.'));
```
ADmad: The use of translation function can be removed. It's highly unlike someone is going to spend time translating messages meant for developers.

src/View/Helper/FormHelper.php
```diff
@@ -41,15 +41,16 @@ class FormHelper extends Helper
      */
     protected $_templates = [
         'dateWidget' => '<ul class="list-inline"><li class="year">{{year}}</li><li class="month">{{month}}</li><li class="day">{{day}}</li><li class="hour">{{hour}}</li><li class="minute">{{minute}}</li><li class="second">{{second}}</li><li class="meridian">{{meridian}}</li></ul>',
-        'error' => '<div class="help-block">{{content}}</div>',
-        'help' => '<div class="help-block">{{content}}</div>',
+        'error' => '<div class="invalid-feedback">{{content}}</div>',
+        'help' => '<small id="{{name}}" class="form-text text-muted">{{content}}</small>',
```
ADmad: Is adding `id` attribute necessary? I don't see any change in code to set the `name` placeholder.

composer.json
```diff
@@ -20,7 +20,9 @@
         "cakephp/cakephp": "~3.0"
```
ADmad: CakePHP version needs to be bumped upto `^3.4`. For e.g. you have used `InstanceConfigTrait::setConfig()` in `PaginatorHelper` which isn't available in older versions.

src/View/Helper/FormHelper.php
```diff
@@ -41,15 +41,18 @@ class FormHelper extends Helper
...
+        'tooltip' => '<span data-toggle="tooltip" title="{{content}}" class="glyphicon glyphicon-info-sign"></span>',
         'inputContainer' => '<div class="form-group {{type}}{{required}}">{{content}}{{help}}</div>',
-        'inputContainerError' => '<div class="form-group {{type}}{{required}} has-error">{{content}}{{error}}{{help}}</div>',
-        'radioInlineFormGroup' => '{{label}}<div class="radio-inline-wrapper">{{input}}</div>',
-        'radioNestingLabel' => '<div class="radio">{{hidden}}<label{{attrs}}>{{input}}{{text}}</label></div>',
-        'staticControl' => '<p class="form-control-static">{{content}}</p>',
+        'inputContainerError' => '<div class="form-group {{type}}{{required}} is-invalid">{{content}}{{error}}{{help}}</div>',
+        'radioInlineFormGroup' => '{{label}}<div class="form-check form-check-inline">{{input}}</div>',
+        'radioNestingLabel' => '<div class="form-check">{{hidden}}<label{{attrs}}>{{input}}{{text}}{{tooltip}}</label></div>',
+        'staticControl' => '<input type="text" readonly class="form-control-plaintext" value="{{content}}>">',
```
ADmad: Static control needs to stay a text element as earlier. As mention in [BS4 migration](https://getbootstrap.com/docs/4.0/migration/#forms-1) `form-control-static` has been renamed to `form-control-plaintext`, so keep the existing template and just change class name.

julianpollmann: Why does it need to be a `<p>`-Tag? Official [Doc's](https://getbootstrap.com/docs/4.0/components/forms/#readonly-plain-text) use the `<input>`-Tag with `form-control-plaintext` class.

ADmad: Using input tag would allow the value to be manipulated through JavaScript. Currently the `staticControl()` generates a P tag and a hidden input with the value, so when using security component the value cannot be manipulated.

src/View/Helper/BreadcrumbsHelper.php
```diff
@@ -10,10 +10,9 @@ class BreadcrumbsHelper extends CoreBreadcrumbsHelper
      */
     protected $_defaultConfig = [
         'templates' => [
-            'wrapper' => '<ol{{attrs}}>{{content}}</ol>',
-            'item' => '<li{{attrs}}><a href="{{url}}"{{innerAttrs}}>{{title}}</a></li>{{separator}}',
-            'itemWithoutLink' => '<li{{attrs}}><span{{innerAttrs}}>{{title}}</span></li>{{separator}}',
-            'separator' => '<li{{attrs}}><span{{innerAttrs}}>{{separator}}</span></li>'
```
ADmad: The core Breadcrumbs helper requires the `separator` template. So please add it back with empty string as template value.

julianpollmann: Fixed with commit 69b9a53

src/Shell/Task/TwbsAssetsTask.php
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Copy sample layouts to app's layout dir
+     *
+     * @param string $target The destination path
+     * @return void
+     */
+    public function copyLayouts($target = null)
+    {
+        $this->info(__('Copying sample layouts...'));
```
ADmad: Translation function should be removed.

src/Template/Bake/Element/form.twig
```diff
@@ -0,0 +1,38 @@
+<div class="{{ pluralVar }} form content">
+    <?= $this->Form->create(${{ singularVar }}) ?>
+    <fieldset>
+        <legend><?= __('{{ action|humanize }} {{ singularHumanName }}') ?></legend>
+        <?php
+{% for field in fields if field not in primaryKey %}
+    {%- if keyFields[field] %}
+        {%- set fieldData = Bake.columnData(field, schema) %}
+        {%- if fieldData.null %}
```
ADmad: Is `fieldData.null` valid usage?

src/Template/Bake/Template/login.twig
```diff
@@ -0,0 +1,15 @@
+<?php $this->extend('../Layout/TwitterBootstrap/signin'); ?>
+
+<?= $this->Form->create(${{ singularVar }}, ['class' => 'form-signin']) ?>
+<img class="mb-4" src="https://getbootstrap.com/assets/brand/bootstrap-solid.svg" alt="" width="72" height="72">
```
ADmad: We shouldn't use external images.

src/Template/Bake/Template/login.twig
```diff
@@ -0,0 +1,15 @@
...
+<h1 class="h3 mb-3 font-weight-normal"><?= __('Please sign in') ?></h1>
+<?= $this->Form->label('email', __('Email address'), ['class' => 'sr-only']) ?>
+<?= $this->Form->email('email', ['placeholder' => __('Email address'), 'autofocus']) ?>
+<?= $this->Form->label('password', __('Password'), ['class' => 'sr-only']) ?>
+<?= $this->Form->password('password', ['placeholder' => __('Password')]) ?>
+<?= $this->Form->control('remember-me', ['type' => 'checkbox']) ?>
+<?= $this->Form->submit(__('Sign in'), ['class' => 'btn btn-lg btn-primary btn-block']) ?>
+<p class="mt-5 mb-3 text-muted">© <?= date('Y') ?></p>
+<?= $this->Form->end() ?>
+
```
ADmad: Empty lines should be removed.

src/Template/Layout/default.ctp
```diff
@@ -7,7 +7,11 @@ use Cake\Core\Configure;
  */
 if (!$this->fetch('html')) {
     $this->start('html');
-    printf('<html lang="%s" class="no-js">', Configure::read('App.language'));
+    if (Configure::check('App.language')) {
+        printf('<html lang="%s">', Configure::read('App.language'));
+    } else {
+        echo('<html>');
```
ADmad: `echo '<html>'`.

src/Template/Layout/default.ctp
```diff
@@ -16,7 +20,9 @@ if (!$this->fetch('html')) {
  */
 if (!$this->fetch('title')) {
     $this->start('title');
-    echo Configure::read('App.title');
+    if (Configure::check('App.title')) {
```
ADmad: This check is unnecessary, if title isn't set nothing is going to be output anyway.

src/View/Helper/BreadcrumbsHelper.php
```diff
@@ -5,34 +5,49 @@
...
+     * Default attributes for the templates
      *
      * @var array
      */
     protected $_defaultAttributes = [
-        'class' => 'breadcrumb'
+        'class' => [
+            'wrapper' => 'breadcrumb',
+            'item' => 'breadcrumb-item',
+            'itemWithoutLink' => 'breadcrumb-item active'
```
ADmad: This doesn't seem to be used anywhere, should it be?

src/View/Helper/HtmlHelper.php
```diff
@@ -5,6 +5,38 @@ class HtmlHelper extends \Cake\View\Helper\HtmlHelper
 {
     use OptionsAwareTrait;
 
+    /**
+     * A list of allowed styles for badges.
+     *
+     * @var array
+     */
+    public $badgeClasses = [
```
ADmad: This array is same as `OptionsAwareTrait::$buttonClasses` so duplicating it under different name doesn't seem DRY. We can just rename `OptionsAwareTrait::$buttonClasses` to a more generic name list `classList` or something and use that everywhere.

src/View/Helper/HtmlHelper.php
```diff
@@ -5,6 +5,38 @@ class HtmlHelper extends \Cake\View\Helper\HtmlHelper
...
+        'light', 'badge-light',
+        'dark', 'badge-dark',
+    ];
+
+    /**
+     * A mapping of aliases for badges.
+     *
+     * @var array
+     */
+    public $badgeClassAliases = [
```
ADmad: Similar comment as above, this array is a duplicate of `OptionsAware::$buttonClassAliases`.

webroot/img/Baked-with-CakePHP.svg
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ADmad: Lower cases file name would be better :)

src/View/Helper/BreadcrumbsHelper.php
julianpollmann: Default templates doesn't include the class anymore. The above classes are injected in the [render](https://github.com/julianpollmann/bootstrap-ui/blob/feature/bootstrap4/src/View/Helper/BreadcrumbsHelper.php#L39) and [add](https://github.com/julianpollmann/bootstrap-ui/blob/feature/bootstrap4/src/View/Helper/BreadcrumbsHelper.php#L49) method.

src/Template/Bake/Element/form.twig
julianpollmann: I'm not an expert in twig. So i took it from [cakephp/bake repo](https://github.com/cakephp/bake/blob/master/src/Template/Bake/Element/form.twig#L51).

ADmad: Hmm I made those templates but don't remember :P

src/View/Helper/HtmlHelper.php
```diff
@@ -72,29 +112,4 @@ public function icon($name, array $options = [])
...
     }
-
-    /**
-     * Returns Bootstrap label markup. By default, uses `<SPAN>`.
-     *
-     * @param string $text Text to show in label.
-     * @param array|string $options Additional HTML attributes.
-     * @return string HTML icon markup.
-     */
-    public function label($text, $options = [])
```
ADmad: There's no alternative in BS4 to generate labels?  FOC/crud-view plugin using this method so I would prefer if this wasn't removed.

ADmad: https://getbootstrap.com/docs/4.0/migration/#labels-and-badges
So it seems "badge" and "label" have been merged. For now be can just keep `HtmlHelper::label()` as alias for `HtmHelper::badge()` for easy migration.

ADmad:
@jotpe I think we can update `HtmlHelper::icon()` and any templates which use glyphicon to instead use font-awesome by default.

ADmad:
@julianpollmann I pushed a commit to use font awesome. :)

src/View/Helper/FormHelper.php
```diff
@@ -44,7 +44,7 @@ class FormHelper extends Helper
         'error' => '<div class="invalid-feedback">{{content}}</div>',
         'label' => '<label{{attrs}}>{{text}}{{tooltip}}</label>',
         'help' => '<small{{attrs}} class="form-text text-muted">{{content}}</small>',
-        'tooltip' => '<span data-toggle="tooltip" title="{{content}}" class="glyphicon glyphicon-info-sign"></span>',
+        'tooltip' => '<span data-toggle="tooltip" title="{{content}}" class="fas fa-info"></span>',
```
dereuromark: fas => fa ?

ADmad: Nope, it's changed in FA 5, check it's docs.

ADmad: If you prefer we can use `info-circle` instead.

dereuromark: Then we should probably mention FA5 requirement vs FA4.x somewhere in the docs to avoid people tripping over it.

CauanCabral: Maybe using "fa fas" should work for both FA 4.7 and 5.

Version 4.7 have more icons free/open - support that version would be great.

ADmad: This template as well as the defaults for `HtmlHelper::icon()` are configurable so those wanting to use FA4 can modify them when loading the helpers.

src/View/UIViewTrait.php
```diff
@@ -29,12 +29,14 @@ public function initializeUI(array $options = [])
...
-        }
+        $helpers = [
+            'Html' => ['className' => 'BootstrapUI.Html'],
+            'Form' => ['className' => 'BootstrapUI.Form'],
+            'Flash' => ['className' => 'BootstrapUI.Flash'],
+            'Paginator' => ['className' => 'BootstrapUI.Paginator'],
+            'Breadcrumbs' => ['className' => 'BootstrapUI.Breadcrumbs'],
+        ];
+
+        $this->helpers = array_merge($helpers, $this->helpers);
```
dereuromark: Should we use assoc + operator? Or do we need to allow recursive merging? I think the first should suffice.

tests/TestCase/View/UIViewTest.php
```diff
@@ -45,4 +45,24 @@ public function testInitialize()
...
+        $View = new UIView(null, null, null, [
+            'helpers' => [
+                'Form' => [
+                    'className' => 'BootstrapUI.Form',
+                    'foo' => 'bar',
+                ],
+            ],
+        ]);
+
+        $expected = ['foo' => 'bar'];
```
dereuromark: looks dead.

src/View/UIViewTrait.php
ADmad: I have intentionally used `array_merge()` to ensure that in the resultant array the keys of `$helpers` stay first. This ensures that if `$this->helpers` contained a helper which for e.g. uses `Html` helper in it's `initialize()` method then that's an instance of `BootstrapUI.HtmlHelper` and not core `HtmlHelper.

tests/TestCase/View/UIViewTest.php
ADmad: Fixed

src/View/Helper/PaginatorHelper.php
```diff
@@ -47,7 +50,7 @@ public function __construct(View $View, array $config = [])
      *   using `'size' => 'lg'` would add class `pagination-lg` to UL tag.
      *
      * @param array $options Options for the numbers.
-     * @return string Numbers string.
+     * @return string|false Numbers string.
      * @link http://book.cakephp.org/3.0/en/views/helpers/paginator.html#creating-page-number-links
      */
     public function numbers(array $options = [])
```
ADmad: Currently `Paginators::numbers()` by default auto adds wrapping class and also has options to generate prev and next links. I don't really like that and would prefer to have the method behavior similar to core method. So I suggest removing these extras and instead add a new method which generates (maybe called `links()`) number links along with wrapper and prev, next, first, last links.

src/View/Helper/FormHelper.php
```diff
@@ -43,7 +43,7 @@ class FormHelper extends Helper
         'dateWidget' => '<ul class="list-inline"><li class="year">{{year}}</li><li class="month">{{month}}</li><li class="day">{{day}}</li><li class="hour">{{hour}}</li><li class="minute">{{minute}}</li><li class="second">{{second}}</li><li class="meridian">{{meridian}}</li></ul>',
         'error' => '<div class="invalid-feedback">{{content}}</div>',
         'label' => '<label{{attrs}}>{{text}}{{tooltip}}</label>',
-        'help' => '<small{{attrs}} class="form-text text-muted">{{content}}</small>',
+        'help' => '<small class="form-text text-muted">{{content}}</small>',
```
dereuromark: why are the attrs removed here? shouldnt that still be possible?

src/View/Helper/FormHelper.php
```diff
@@ -285,20 +285,10 @@ public function control($fieldName, array $options = [])
         }
 
         if ($options['help']) {
-            if (is_string($options['help'])) {
-                $options['help'] = $this->templater()->format(
-                    'help',
-                    ['content' => $options['help']]
-                );
-            } elseif (is_array($options['help'])) {
```
ADmad: Why did you remove the ability to set 'help' key to an array? This feature was recently added on master as someone wanted to add extra attributes to the tag used for help message.

src/View/Helper/FormHelper.php
julianpollmann: Setting the attrs (e.g. class) will end up in having duplicated class attributes.
E.g. this
`$this->Form->control('email', ['help' => ['class' => 'foo', 'content' => 'someHelp']]);` will end up with
`<small class="foo" content="someHelp" class="form-text text-muted">someHelp</small>`

Since help itself is added as a option to a control the injectClasses() method seems not to work here?!

ADmad: People can set other attributes besides class, for e.g. "id".

src/Shell/BootstrapShell.php
```diff
@@ -43,11 +43,15 @@ public function copyLayouts()
     /**
      * Modifies AppView to extend UIView
      *
+     * @param tring $path Path to AppView
```
dereuromark: string|null

ADmad:
Test failing for BootstrapShell.

dereuromark:
I tested it and it works nicely.
The only thing I couldnt get to work was the fade in flash messages. I had to 
```
.fade {
    opacity: inherit;
}
```
here.

We could also set up an updated v4 version @ https://sandbox.dereuromark.de/sandbox/bootstrap
Show case all things in real time / life.

ADmad:
@dereuromark Having a demo on your sandbox for v4 sounds great. That would help us test and fix any remaining issues. Can you setup a demo using the `develop` branch?

src/Shell/Task/TwbsAssetsTask.php
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Installs Bootstrap assets using npm
+     *
+     * @return void
+     */
+    public function installAssets()
+    {
+        $this->info('Checking npm...');
+        if (!`which npm`) {
```
ADmad: Shouldn't there be an `exec()` call here?

ADmad: NVM, I missed the backticks.

ADmad:
The test for `BootstrapShell` most likely fails because the `TwbsAssets` task bails out if `npm` is not found. So you need to add a check to skip the test too if `npm` is not available.

ADmad:
Merging so that people can start trying out the `develop` branch. Hopefully someone will contribute a fix for the tests issue I pointed out above soon.

PR #215, opened 2018-04-21
change deprecated methods

cewi:
## Description
changes some deprecated methods
minimum cakephp version is now 3.5 (not sure if necessary)


cewi:
ok, but develop is for bootstrap4 and master for bootstrap3. So there will be seperate branches in future? Looking forward -,)

ADmad:
@cewi Please reopen the PR :) Master won't be merged into develop, due to difference in bootstrap version it would create lot of conflicts.

dereuromark:
I also said "might" - and "some", not "all" :)

cewi:
no problem. HTH

ADmad:
@cewi You need to update the deprecations in test cases too :slightly_smiling_face:.

cewi:
@ADmad: There a still a lot of tests against Form->input() method in FormHelperTest.php. They do not throw deprecation errors. Should they be updated anyway?

ADmad:
@cewi Yes please.

tests/TestCase/View/Helper/FormHelperTest.php
```diff
@@ -81,28 +91,11 @@ public function tearDown()
         I18n::setLocale($this->locale);
     }
 
-    public function testBasicTextInput()
```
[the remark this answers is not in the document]
cewi: yes exactly

tests/TestCase/View/Helper/FlashHelperTest.php
```diff
@@ -4,8 +4,8 @@
 
 use BootstrapUI\View\Helper\FlashHelper;
 use Cake\Core\Exception\Exception;
-use Cake\Network\Request;
-use Cake\Network\Session;
+use Cake\Http\ServerRequest;
+use Cake\Http\Session;
```
ADmad: This change causes tests to fail on 3.5 and we are trying to maintain 3.5 compatibility too. So to make this work on both 3.5 and 3.6 you remove the `use` statement for session class and instead inside `setUp()` use a `class_exists()` check and accordingly use appropriate class to create instance.

cewi:
did you mean like so?

tests/TestCase/View/Helper/FlashHelperTest.php
```diff
@@ -34,8 +33,17 @@ public function setUp()
     {
         parent::setUp();
         $this->View = new View();
-        $session = new Session();
-        $this->View->request = new Request(['session' => $session]);
+
+        // load appropriate session class according to Cake version
+        if (!class_exists('\Cake\Http\Session')) {
```
ADmad: You can remove the negation and swap the `if` and `else` parts :slightly_smiling_face:

cewi: yeah, this complexity is really useless

ADmad:
Thanks!

PR #217, opened 2018-05-18
Fixing detectFormAligment method

viniciusbig:
## Description
There was a small bug in the detectFormAligment method. 
checkClasses() uses the entire $options array and not just only the 'class' part

## Summarize
- Change the parameter on checkClasses call

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -549,7 +549,7 @@ class FormHelper extends Helper
     protected function _detectFormAlignment($options)
     {
         foreach (['horizontal', 'inline'] as $align) {
-            if ($this->checkClasses('form-' . $align, (array)$options['class'])) {
+            if ($this->checkClasses('form-' . $align, $options)) {
                 return $align;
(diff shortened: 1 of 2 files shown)


dereuromark:
This seems to be missing a test case.

dereuromark:
Timing^^

dereuromark:
You can also send as the HTML of expected vs is and we can try to add code to your PR.
It would be best if you could do the test case on your own though :)

dereuromark:
ping @viniciusbig

viniciusbig:
Hello guys!

I was reading and studying a lot PHP unit test in the last weeks. Sorry for a long time without an update.

In the end, I've just added a simple test case to confirm that `class` key is present on `checkClasses` parameters.

I'm not an expert on this, but I learned a lot during this time!

Can you check the PR, please?

dereuromark:
Thank you :) Looks good to me.

dereuromark:
But why did the other tests before work? If only the classes key was passed?

Issue #114, opened 2016-01-22
RFC, Depending on Bootstrap


davidyell:
As there is a repo for it on Packagist.
https://packagist.org/packages/twbs/bootstrap

Should the plugin either `require` it, or `suggest` it and provide a shell task to symlink the resources into the `webroot/css` folder?

ADmad:
It's pretty uncommon to manage js / css assets through composer.

ayman-alkom:
what about having a bower.json and set the bower component folder to webroot/assets and make en element for the stuff of including them.
I think that every developer has the same issue with initializing the plugin

davidyell:
My problem with Bower is that it comes with a number of dependencies. Bower requires NPM, which requires Node. It would certainly help anyone using Bower already for front-end assets though for sure.

I just feel that the plugin could do with *something* in this area, as including Bootstrap in your plugin is pretty key to making this plugin 'work' in the intended fashion. That might be a Composer suggest, or even just some updates to the readme to appraise users of the possibilities. Whether that be Bowser, Composer or Zip file.

ADmad:
Providing a "suggest" for composer won't do much unless the plugin also provides instructions/shell to copy/link the assets under webroot.

ayman-alkom:
yes @davidyell  you are right about the dependencies of bower, but I meant that we can keep the assets at the webroot, so we dont need to run bower unless who want to update the assets will need the node and NPM

jadb:
I will document how I use it with composer on some setups as a proposed RFC and we shall see what the majority thinks.

/cc @deizel

ADmad:
v2 of the plugin provides a shell which install assets using `npm`. While BS4 and jquery are available through `composer` BS4 also requires `popper.js` which isn't, so it's not feasible to use composer to get install all assets too.

PR #220, opened 2018-05-22
use composer in favor of bower

Rubyan:
Not everyone has bower installed, and bower itself seems to be [deprecated]:https://snyk.io/blog/bower-is-dead/ . Since most users are already using composer, they can use it to install the correct bootstrap assets too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,21 +142,19 @@ Available types are:
 
 **NOTE: Remember to set the stylesheets in the layouts you copy.**
 
-## Installing Bootstrap via Bower
+## Installing Bootstrap via Composer
 
-A quick way of getting the Bootstrap assets installed is using [bower]. Assuming you are in `ROOT`:
+A quick way of getting the Bootstrap and jQuery assets installed is using composer. Assuming you are in `ROOT`:
 
 ```
-bower install bootstrap
-bower install jquery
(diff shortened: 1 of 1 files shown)


davidyell:
Hah, I proposed this in #114

dereuromark:
:-1: I am against using composer for frontend asset management and mixing those with php dependencies.
There are tools to do that, and composer probably shouldnt be the default one.

I would be OK to just add a paragraph at the end noting the composer usage on top, but the standard IMO should not be changed here.

ADmad:
Showing use of composer is better than showing use of a defunkt js based tool.

Either way devs are free to use whatever they want to get the assets and the plugin doesn't force you to use any particular one.

ADmad:
Any PR to showing/mentioning additional ways to acquire BS3 and jQuery are welcome.

sdevore:
I've been using https://asset-packagist.org in a new project that I am just ramping up.  I do think that before I go to actual deployment I will probably split out the npm/bower dependancies.  But it has been working smoothly for me at this point.

ADmad:
@sdevore That's interesting, I'll check it out.

davidyell:
I know of a few,

- https://unpkg.com/
- Good old `npm install` or `yarn add`

PR #221, opened 2018-06-03
Fix bootstrap shell install test failure.

ndm2:
By default the `js` folder doesn't exist in the plugins webroot folder,
and the task doesn't create it, hence `$this->_jsDir->path` will be
`null`, causing the assets to be copied in the wrong directory.

refs #207

I've set all required folders to be auto-created, just in case existing folders might be removed in the future.

### DIFF
--- a/src/Shell/BootstrapShell.php
+++ b/src/Shell/BootstrapShell.php
@@ -27,7 +27,8 @@ class BootstrapShell extends Shell
     {
         $this->TwbsAssets->installAssets();
         $this->TwbsAssets->copyAssets();
-        $this->Assets->symlink();
+        $this->Assets->remove('BootstrapUI');
+        $this->Assets->symlink('BootstrapUI');
     }
 
     /**
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "cakephp/cakephp": "^3.5"
+        "cakephp/cakephp": "^3.5.12"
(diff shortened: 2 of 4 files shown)


src/Shell/BootstrapShell.php
```diff
@@ -27,7 +27,8 @@ public function install()
     {
         $this->TwbsAssets->installAssets();
         $this->TwbsAssets->copyAssets();
-        $this->Assets->symlink();
+        $this->Assets->remove('BootstrapUI');
```
ADmad: `AssetsTask::remove()` was added in `3.5.12`, so please update `composer.json` accordingly.

ndm2:
Shouldn't `^4.0.0` include the latest `4.1`?

ADmad:
Right, it would. So no change necessary :)

Btw have you tried out v2 in an actual app? In your opinion can we goto RC or stable release?

ndm2:
Not yet, but I'm working on a project that will require Bootstrap 4. The work on the app should start some time in the next two weeks. I'll report back on my findings then :)

ADmad:
Great, would love to have your feedback. Will just release another beta in the mean time :slightly_smiling_face:.

ndm2:
I think you've tagged the wrong branch ;)

ADmad:
Oops, I'll fix it soon.

ADmad:
Fixed :blush:

Issue #223, opened 2018-06-18
Various Bootstrap 4 variant issues/suggestions


ndm2:
I've only just begun using the plugin in a new project, but I wanted to open an issue already to keep track of my findings, especially since I don't know exactly how long it will take for the project to finish.

I'll extend the following list over time and will report when I'm done.

* [ ] Labels get the `col-form-label` class injected for non-horizontal forms.

* [ ] Baked `add.ctp` and `edit.ctp` files seem to miss newlines in the `tb_actions` block (before and after actions for associations to be specific).

* [ ] Maybe submit buttons (those created via `FormHelper::submit()`) could use the `btn-primary` class by default. At least personally I've found myself to always switch to that.

ndm2:
@ADmad Are you (or anyone else for that matter) already working on the nested input problem? If not, I'll give it a shot.

ADmad:
@ndm2 I am not and pretty sure no one else is either :slightly_smiling_face:.

ndm2:
Yeah, I had a quick look at it, and I can totally understand that no on wants to touch it :slightly_frowning_face: Well, I'll give it a try, maybe it's going to be less of a hassle than it looks like.

ndm2:
Actually, it's even more hassel-ish. Cakes internal handling of nested inputs seems to make it more complicated than it needs to be, and I'm not even sure if CakePHP is doing things "correctly" here.

When using `'nestedInput' => true` for a radio control, I would expect to get the default HTML, ie as if `nestedInput` isn't set, but instead it generates weird HTML like this:

```html
<div class="input radio">
	<label>
		<input type="hidden" name="Radio (nested)" value="">
		<label for="radio-nested-1">
			<input type="radio" name="Radio (nested)" value="1" id="radio-nested-1">
			foo
		</label>
		<label for="radio-nested-2">
			<input type="radio" name="Radio (nested)" value="2" id="radio-nested-2">
			bar
		</label>
		<label for="radio-nested-3">
			<input type="radio" name="Radio (nested)" value="3" id="radio-nested-3">
			baz
		</label>
		Radio (nested)
	</label>
</div>
```

where the nested radios are using the `radioWrapper` as well as `nestingLabel` templates, and the wrapping `<label>` uses the `nestingLabel` template too.

That's ultra double plus non-helpful :grimacing:

ADmad:
The wackiness in core could be fixed in 4.0 if you can make a PR.

[...]

PR #225, opened 2018-07-08
Develop fixes

ADmad:
Refs #223.

Following issues are fixes:

* Labels get the `col-form-label` class injected for non-horizontal forms.
* Checkbox wrappers in non-horizontal forms are missing the `is-invalid` class.
* Horizontal forms are missing the `row` class on the control wrapper `div`.
* Horizontal forms use a column width of `6` for inputs, the default should be `10`. The `right` configuration of the `grid` option that eats the remaining column width of `4`, is never used.
* The offset class used in horizontal forms uses the Bootstrap 3 name `col-%-offset-%`, it should be `offset-%-%` instead.
* `AppView` class in the test app pollutes the App namespace (can confuse IDEs).

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -522,7 +522,7 @@ class FormHelper extends Helper
     {
         $class = 'col-%s-';
         if ($offset) {
-            $class .= 'offset-';
+            $class = 'offset-%s-';
         }
 
         if (isset($this->_grid[$position])) {
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -79,8 +79,8 @@ class FormHelper extends Helper
             'inputContainerError' => '<div class="form-group {{type}}{{required}} is-invalid">{{content}}</div>',
             'checkboxContainer' => '<div class="form-group {{type}}{{required}}">{{content}}</div>',
             'checkboxContainerError' => '<div class="form-group {{type}}{{required}} is-invalid">{{content}}</div>',
-            'radioContainer' => '<div class="form-group{{required}}">{{content}}</div>',
-            'radioContainerError' => '<div class="form-group{{required}} is-invalid">{{content}}</div>',
+            'radioContainer' => '<div class="form-group {{type}}{{required}}">{{content}}</div>',
+            'radioContainerError' => '<div class="form-group {{type}}{{required}} is-invalid">{{content}}</div>',
         ]
     ];
 
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -63,8 +63,8 @@ class FormHelper extends Helper
      */
     protected $_templateSet = [
         'default' => [
-            'checkboxContainer' => '<div class="form-check"{{required}}>{{content}}{{help}}</div>',
-            'checkboxContainerError' => '<div class="form-check">{{content}}{{error}}{{help}}</div>',
+            'checkboxContainer' => '<div class="form-group {{type}}{{required}}">{{content}}{{help}}</div>',
+            'checkboxContainerError' => '<div class="form-group {{type}}{{required}} is-invalid">{{content}}{{error}}{{help}}</div>',
         ],
         'inline' => [
             'label' => '<label class="sr-only"{{attrs}}>{{text}}{{tooltip}}</label>',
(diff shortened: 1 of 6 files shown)


ADmad:
The points listed in #223 regarding `nestedInput` option and `form-check-label` / `form-check-input`, `form-check` class are kinda all related.

Bootstrap 4 has got rid of the hackery used in v3 of nesting radio/checkbox inputs inside label. Properly applying the classes mention above will actually require to turning off the default nesting that core's `FormHelper` does and adjust BUI's templates/code accordingly.

ndm2:
> Properly applying the classes mention above will actually require to turning off the default nesting [...]

Maybe that's what should be done? While the nested variant works, as in, the checkbox will be usable, it's not being displayed properly, and requires to use "wrong" wrapper classes, and it may also interfer with Bootstrap extensions that expect the non-nested elements.

Not a fan of how CakePHP handles this internally. It looks like it's untested, and it seems that one must also modify the `checkboxFormGroup` template additionally to disabling `nestedInput`, and the option seems to have no effect for `radio` controls.

ADmad:
> Maybe that's what should be done? While the nested variant works, as in, the checkbox will be usable, it's not being displayed properly, and requires to use "wrong" wrapper classes so that the alignment isn't off, and it may also interfer with Bootstrap extensions that expect the non-nested elements.

Yup, I also noticed that the alignment if off with current nesting. So I will work on removing the default nesting, unless someone beats me to it.

> Not a fan of how CakePHP handles this internally. It looks like it's untested, and it seems that one must also modify the checkboxFormGroup template additionally to disabling nestedInput, and the option seems to have no effect for radio controls.

The default nesting done in core is also as a result of numerous issues/requests by bootstrap 3 users. Now that bootstrap has finally removed it we can change in core too, since to the best of my knowledge no other UI library does that. But that would probably have to wait till Cake 5.

src/View/Helper/FormHelper.php
```diff
@@ -75,12 +75,12 @@ class FormHelper extends Helper
             'formGroup' => '{{label}}<div class="%s">{{input}}{{error}}{{help}}</div>',
             'checkboxFormGroup' => '<div class="%s"><div class="checkbox">{{label}}</div>{{error}}{{help}}</div>',
             'submitContainer' => '<div class="form-group"><div class="%s">{{content}}</div></div>',
```
ndm2: This should have `row` too I think.

src/View/Helper/FormHelper.php
```diff
@@ -63,8 +63,8 @@ class FormHelper extends Helper
      */
     protected $_templateSet = [
         'default' => [
-            'checkboxContainer' => '<div class="form-check"{{required}}>{{content}}{{help}}</div>',
-            'checkboxContainerError' => '<div class="form-check">{{content}}{{error}}{{help}}</div>',
+            'checkboxContainer' => '<div class="form-group {{type}}{{required}}">{{content}}{{help}}</div>',
```
ndm2: This should stay `form-check`, but it's related to the nesting problem, so, guess that can be addressed later on.

ndm2:
Except for the two comments, it seems that this fixes the mentioned issues. Though I've stumbled over an additional `col-form-label` problem, that is that it needs an additonal `pt-0` when it's in a horizontal form and for a checkbox (or radio button) group.

Check https://getbootstrap.com/docs/4.1/components/forms/#horizontal-form, without the `pt-0` the vertical alignment will be off. Also just noticed that the label shouldn't be a `label` in this case, but a `legend`, and the wrapper should be a `fieldset`, but that's a different problem :)

src/View/Helper/FormHelper.php
ADmad: Yes, it's related to the nesting so haven't made any change related to that.

src/View/Helper/FormHelper.php
ADmad: I haven't made any related to your issue with `FormHelper::submit()`. Why are you even using submit inputs instead of generating buttons using `FormHelper::button()`? :slightly_smiling_face:

ADmad:
> Except for the two comments, it seems that this fixes the mentioned issues. Though I've stumbled over an additional col-form-label problem, that is that it needs an additonal pt-0 when it's in a horizontal form and for a checkbox (or radio button) group.

I have intentionally overlooked details regarding checkbox and radio generation. Those will have to be taken care of along with the removal of input nesting.

> Check https://getbootstrap.com/docs/4.1/components/forms/#horizontal-form, without the pt-0 the vertical alignment will be off. Also just noticed that the label shouldn't be a label in this case, but a legend, and the wrapper should be a fieldset, but that's a different "problem" :)

F***, that's going to need lot of changes.

src/View/Helper/FormHelper.php
ndm2: Old habits I guess, and the alignment with the inputs was visually more appealing :)

PR #227, opened 2018-07-10
Issue/223 - Breadcrumbs Helper

ndm2:
Adds missing overrides to ensure bootstrap classes are added for all
crumb insertion methods.

Ensures that the `aria-current` attribute is only applied to one crumb,
either the last one (default), or the last one with a link, and that for
crumbs with links the attribute is set on the link.

refs #223

### DIFF
--- a/src/View/Helper/BreadcrumbsHelper.php
+++ b/src/View/Helper/BreadcrumbsHelper.php
@@ -6,14 +6,16 @@ use Cake\View\Helper\BreadcrumbsHelper as CoreBreadcrumbsHelper;
 class BreadcrumbsHelper extends CoreBreadcrumbsHelper
 {
     use OptionsAwareTrait;
+
     /**
      * @inheritdoc
      */
     protected $_defaultConfig = [
+        'ariaCurrent' => 'last',
         'templates' => [
             'wrapper' => '<nav aria-label="breadcrumb"><ol{{attrs}}>{{content}}</ol></nav>',
             'item' => '<li{{attrs}}><a href="{{url}}"{{innerAttrs}}>{{title}}</a></li>',
-            'itemWithoutLink' => '<li{{attrs}} aria-current="page"><span{{innerAttrs}}>{{title}}</span></li>',
+            'itemWithoutLink' => '<li{{attrs}}><span{{innerAttrs}}>{{title}}</span></li>',
             'separator' => ''
         ]
     ];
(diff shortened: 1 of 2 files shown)


src/View/Helper/BreadcrumbsHelper.php
```diff
@@ -50,4 +53,70 @@ public function add($title, $url = null, array $options = [])
...
+    }
+
+    /**
+     * Injects the `aria-current` attribute into the current set of crumbs.
+     *
+     * @return void
+     */
+    protected function _injectAriaCurrentAttribute()
+    {
+        if ($this->crumbs) {
```
ADmad: Can you change this to "early return" please.

src/View/Helper/BreadcrumbsHelper.php
```diff
@@ -50,4 +53,70 @@ public function add($title, $url = null, array $options = [])
...
+                foreach (array_reverse($this->crumbs, true) as $key => $crumb) {
+                    if (isset($crumb['url'])) {
+                        break;
+                    }
+                }
+            } else {
+                $key = count($this->crumbs) - 1;
+            }
+
+            if ($key) {
```
ADmad: "early return" here too.

tests/TestCase/View/Helper/BreadcrumbsHelperTest.php
```diff
@@ -22,9 +23,9 @@ public function setUp()
 
         $this->View = new View();
         if (class_exists('\Cake\View\Helper\BreadcrumbsHelper')) {
```
ADmad: The `develop` branch is for Cake `3.5+` which has the `BreadcrumbsHelper` hence this check can be removed.

tests/TestCase/View/Helper/BreadcrumbsHelperTest.php
```diff
@@ -12,7 +13,7 @@ class BreadcrumbsHelperTest extends TestCase
     public $View;
 
     /**
-     * @var HtmlHelper
+     * @var BreadcrumbsHelper
```
ADmad: Nitpick: Needs to be FQCN to be actually useful to IDEs :slightly_smiling_face:

ndm2: And the non-PhpStorm peasants should suffer, for they have sinned :grimacing: 

I can change it, but none of the the other `@var` tags in the test files uses fully qualified names. I can change them all if you want?

dereuromark: You could just use my Spryker sniffer which fixes it for you globally :P

ADmad: It's a test file so I don't really care :P

PR #232, opened 2018-08-23
Get always minified assets.

snize:
## Description
Changed to get minified css and scripts in `default.ctp` by default.
If debug is false `bin/cake bootstrap install` installs assets only minified one.
But templates still render non-minified file path, So gets 404 error with those.


ADmad:
Closing in favor of #233.

PR #233, opened 2018-08-23
Get minified assets if debug is false.

snize:
## Description
Almost same as #232 but load minified assets if debug is false.

Still gets 404 if you switch debug false to true after `bin/cake bootstrap install`.


ADmad:
It would be better to also modify `TwbsAssetsTask::copyAssets()` to copy minified as well as unminified files so that turning debug on/off does not create problems.

dereuromark:
Looks good.

PR #234, opened 2018-10-04
Update FormHelper.php

highstrike:
Refs #223 

So i'm trying bootstrap 4 with cake and stumbled upon the following problem:
- Labels/inputs for checkboxes and radiobuttons are missing the `form-check-label/form-check-input` class.

I made a quick and dirty solution for checkboxes which works for me.
It adds the proper classes and moves the input outside of the label without having to mess with `nestedInput`.

Maybe the same could be done for radios, but I haven't touched those yet.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -63,8 +63,8 @@ class FormHelper extends Helper
      */
     protected $_templateSet = [
         'default' => [
-            'checkboxContainer' => '<div class="form-group {{type}}{{required}}">{{content}}{{help}}</div>',
-            'checkboxContainerError' => '<div class="form-group {{type}}{{required}} is-invalid">{{content}}{{error}}{{help}}</div>',
+            'checkboxContainer' => '<div class="form-group form-check {{type}}{{required}}">{{content}}{{help}}</div>',
+            'checkboxContainerError' => '<div class="form-group form-check {{type}}{{required}} is-invalid">{{content}}{{error}}{{help}}</div>',
         ],
         'inline' => [
             'label' => '<label class="sr-only"{{attrs}}>{{text}}{{tooltip}}</label>',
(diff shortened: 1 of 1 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -253,6 +253,11 @@ public function control($fieldName, array $options = [])
                         $options['templates']['checkboxContainer'] = '{{content}}';
                     }
                 }
+
+                $options['templates']['nestingLabel'] = '{{hidden}}{{input}}<label{{attrs}}>{{text}}</label>';
```
ADmad: Since BS4 doesn't need nested inputs for checkbox and radio instead of setting this template it would be better to modify relevant code to simply not use nesting labels.

highstrike: if `nestedInput` is false, the input doesn't appear at all, same issue that @ndm2 had

ADmad: Well than let's fix the cause rather than the symptom. :slightly_smiling_face:

highstrike: isn't that a core problem? one which might someday get fixed in cake4? relying on nesting labels for now and updating the template to put the input outside is at least a workaround until then

ADmad: BUI's form helper overrides core's so we can make the necessary fix here. Once I understand the source of the problem I can fix core too.

ndm2:
That's similar to what I've done in my yet to be completed changes (sorry, have to work on three projecst simultaneously currently, I'm practically ded :)).

I've opted for making non-nested inputs the default, and using separate templates to still support nesting (surprinsingly there's quite some BS4 extensions out there that rely on the nested variant), and let the user change the markup.

ADmad:
Thanks @highstrike, I didn't notice you had updated the PR. 

But closing in favor of #235 as I have updated tests too there and fixed radio too. :slightly_smiling_face:

PR #236, opened 2018-10-05
Make submit button 'primary' by default

highstrike:
Refs #223 

- [x] Maybe submit buttons (those created via FormHelper::submit()) could use the btn-primary class by default. At least personally I've found myself to always switch to that.

ping @ndm2

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -189,6 +189,9 @@ class FormHelper extends Helper
      */
     public function submit($caption = null, array $options = [])
     {
+        $options += [
+            'class' => 'primary'
+        ];
         $options = $this->applyButtonClasses($options);
 
         return parent::submit($caption, $options);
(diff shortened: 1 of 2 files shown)


highstrike:
i don't understand why those tests fail...

ADmad:
Please try to use proper commit messages, "Update FormHelperTest.php" is not good enough :)

highstrike:
the last one was a revert to a comma to test travis

ADmad:
Then "Fix CS errors" would be a more useful message.

ADmad:
Btw why are people still using submit inputs instead of buttons?

highstrike:
okay, but the point is that i still don't know why travis failed

highstrike:
as for buttons over submit inputs, take it up with upper management 😁  https://github.com/cakephp/cakephp/blob/master/src/View/Helper/FormHelper.php#L135

ADmad:
> but the point is that i still don't know why travis failed

Debug the html generated in side test case and find out :)

PR #239, opened 2018-11-02
Use newer, non-deprecated method

jonlachmann:
Use the newer getParam instead of older direct access of controller from view.

## Description
Use the newer getParam instead of older direct access of controller from view.

## Summarize
Just changed the two occurences in the file default.ctp

## Benefits
No deprecation warnings when changed.

### DIFF
--- a/src/Template/Layout/default.ctp
+++ b/src/Template/Layout/default.ctp
@@ -32,7 +32,7 @@ if (!$this->fetch('tb_footer')) {
 /**
  * Default `body` block.
  */
-$this->prepend('tb_body_attrs', ' class="' . implode(' ', [$this->request->controller, $this->request->action]) . '" ');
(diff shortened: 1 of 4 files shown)


ADmad:
In the related issue you mentioned there were more such occurrences?

dereuromark:
Can you please also fix the missing brackets around 

    if (isset($this->Flash))

seems like someone coded a bit dirty and the code sniffer doesnt find these things.

jonlachmann:
> Can you please also fix the missing brackets around
> 
> ```
> if (isset($this->Flash))
> ```
> seems like someone coded a bit dirty and the code sniffer doesnt find these things.

How should it be? I am happy to add it in the pull request, if I can figure out how, but I'd rather do it right than wrong.

dereuromark:
With brackets like the other if code. Without brackets is not allowed as per cs/psr2.

jonlachmann:
Now I have fixed all occurrences of the deprecated method and also added the brackets, and managed to put them in the right branch to add them to the pull request.

PR #244, opened 2018-11-30
BS4 Compatibility - Multi Checkbox Control

ndm2:
Instead of waiting till I can get everything working, I thought I'd push things out incrementally per control, this has been on the back burner for way too long already :/

Nested inputs for multi option controls required the changes in #242 to be altered a bit, in order to avoid duplicate inputs, and to retain the ability the disable labels per option configuration - hopefully this won't come back to bite me with the remaining controls.

[**Errors aren't displayed properly**](https://github.com/FriendsOfCake/bootstrap-ui/issues/223#issuecomment-405334736) yet, but this is also true for other controls, so I'm going to work on that later on when all controls are rendering the correct markup.

For reference

* https://getbootstrap.com/docs/4.1/components/forms/#horizontal-form
* https://getbootstrap.com/docs/4.1/components/forms/#inline
* #223

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -60,9 +60,11 @@ class FormHelper extends Helper
         'inputGroupContainer' => '<div{{attrs}}>{{prepend}}{{content}}{{append}}</div>',
         'inputGroupText' => '<span class="input-group-text">{{content}}</span>',
         'file' => '<input type="file" class="form-control-file" name="{{name}}"{{attrs}}>',
-        'multicheckboxContainer' => '<fieldset class="form-group {{type}}{{required}}">{{content}}{{help}}</fieldset>',
-        'multicheckboxContainerError' => '<fieldset class="form-group {{type}}{{required}} is-invalid">{{content}}{{error}}{{help}}</fieldset>',
-        'multicheckboxLegend' => '<legend>{{text}}</legend>',
+        'multicheckboxContainer' => '<div class="form-group {{type}}{{required}}" role="group" aria-labelledby="{{groupId}}">{{content}}{{help}}</div>',
+        'multicheckboxContainerError' => '<div class="form-group {{type}}{{required}} is-invalid" role="group" aria-labelledby="{{groupId}}">{{content}}{{error}}{{help}}</div>',
+        'multicheckboxLabel' => '<label id="{{groupId}}" class="d-block">{{text}}</label>',
+        'multicheckboxWrapper' => '<fieldset class="form-group">{{content}}</fieldset>',
+        'multicheckboxTitle' => '<legend class="col-form-label pt-0">{{text}}</legend>',
         'nestingLabel' => '{{hidden}}{{input}}<label{{attrs}}>{{text}}{{tooltip}}</label>',
         'nestingLabelNestedInput' => '{{hidden}}<label{{attrs}}>{{input}}{{text}}{{tooltip}}</label>',
     ];
@@ -78,9 +80,9 @@ class FormHelper extends Helper
         'inline' => [
             'label' => '<label class="sr-only"{{attrs}}>{{text}}{{tooltip}}</label>',
             'inputContainer' => '{{content}}',
-            'multicheckboxContainer' => '<div class="form-group {{type}}{{required}}">{{content}}</div>',
-            'multicheckboxContainerError' => '<div class="form-group {{type}}{{required}} is-invalid">{{content}}</div>',
-            'multicheckboxLegend' => '<span class="sr-only">{{text}}</span>',
+            'multicheckboxContainer' => '<div class="form-group {{type}}{{required}}" role="group" aria-labelledby="{{groupId}}">{{content}}</div>',
+            'multicheckboxContainerError' => '<div class="form-group {{type}}{{required}} is-invalid" role="group" aria-labelledby="{{groupId}}">{{content}}</div>',
+            'multicheckboxLabel' => '<span id="{{groupId}}" class="sr-only">{{text}}</span>',
         ],
         'horizontal' => [
             'label' => '<label class="col-form-label %s"{{attrs}}>{{text}}{{tooltip}}</label>',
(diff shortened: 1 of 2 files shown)


ADmad:
> in order to avoid duplicate inputs

I thought I had taken care of duplicates, guess not.

Reading test cases for generated HTML is a PITA. We need a demo app to see the generated output :slightly_smiling_face:.

ndm2:
Yeah, it's not pretty :) I have a simple app that I'm using for visualization and markup inspection purposes, I'll upload it later today.

ndm2:
Here you go, it's very basic, it simply lists a bunch of controls per alignment, and shows the code alongside of it. The code formatting validates (and unfortunately fixes) the HTML too, and will output warnings accordingly, the original code can be viewed in the "Raw" tabs.

https://github.com/ndm2/bootstrap-ui-test-app

src/View/Helper/FormHelper.php
```diff
@@ -306,6 +324,20 @@ public function control($fieldName, array $options = [])
                         $options['label'] = $this->injectClasses('form-check-label', (array)$options['label']);
                     }
                 }
+
+                if ($options['type'] === 'multicheckbox') {
+                    $options['templates']['label'] = $this->templater()->get('multicheckboxLegend');
```
ADmad: Core uses `label` tag for the common title for multi checkboxes, why do we need to change it to legend?

ndm2: Are you referring to the generated markup, or the template name? The `legend` markup is the correct thing to use because the wrapper is a `fieldset`. And while the bootstrap docs are using `fieldset` for radios only, I think using it for multi checkboxes makes sense too.

ADmad: I am referring to the generated markup. If I comment this whole `if` block the markup generated looks good enough with the `<label>` tag. While using `<legend>` might be semantically correct I wonder if deviating from core is necessary when there are no display issues.

src/View/Helper/FormHelper.php
```diff
@@ -50,15 +50,21 @@ class FormHelper extends Helper
...
+        'checkboxInlineWrapper' => '<div class="form-check form-check-inline">{{label}}</div>',
+        'radioWrapper' => '<div class="form-check">{{hidden}}{{label}}</div>',
         'radioInlineFormGroup' => '{{label}}<div class="form-check form-check-inline">{{input}}</div>',
         'radioNestingLabel' => '<div class="form-check">{{hidden}}<label{{attrs}}>{{input}}{{text}}{{tooltip}}</label></div>',
         'staticControl' => '<p class="form-control-plaintext">{{content}}</p>',
         'inputGroupAddon' => '<div class="{{class}}">{{content}}</div>',
         'inputGroupContainer' => '<div{{attrs}}>{{prepend}}{{content}}{{append}}</div>',
         'inputGroupText' => '<span class="input-group-text">{{content}}</span>',
-        'file' => '<input type="file" class="form-control-file" name="{{name}}"{{attrs}}>'
+        'file' => '<input type="file" class="form-control-file" name="{{name}}"{{attrs}}>',
```
ADmad: Not related to this PR which only deals with multicheckbox but we need to fix the template for `file`. As seen in your demo app currently the input tag has the `class` attribute twice.

src/View/Helper/FormHelper.php
ADmad: Okay, display of "select (checkbox, inline)" does have a slight problem.

ndm2: The same would apply when switching radios to bootstraps suggested markup.

I guess from an accesibility perspective, `label` would work in a `fieldset` too, and it would be possible to for example add a `d-block` class to fix the layout issue. `legend` however is already included in the bootstrap CSS and the docs suggest its use. Then again, in a non-horizontal layout, the `legend` looks off, which however could be fixed by applying the `col-form-label pt-0` classes that are already used for horizontal align.

src/View/Helper/FormHelper.php
ndm2: Indeed. That will then also fix the missing error message.

src/View/Helper/FormHelper.php
ADmad: For horizontal form (`/pages/horizontal-align#select-checkbox-anchor`), with your change we end up with `<legend>` tag nested inside inside `<div>` tag. That's invalid markup. 

`validator.w3.org` says `Element legend not allowed as child of element div in this context.`

ndm2: Yeah, the specs are pretty resrictive there (for no good reason IMHO), the problem is that styling won't work properly when using the "valid" markup.

https://github.com/twbs/bootstrap/issues/24938

I'm not married to the idea of using fieldsets, I think it would be nice if we could get it working, but I'd be fine with using divs and labels for now if they have a better pros/cons ratio.

ADmad: Reading that issue I fail to see why they can't achieve same display while still having valid HTML but let's not get into that.

Given that core uses `<label>` tag by default and using `<legend>` generates invalid markup I would rather stick with what core markup and see if we can just fix the display issue by applying required classes to the `<label>` tag.

I tried checking what other frameworks do and found this e.g. of [horizontal form](https://bulma.io/documentation/form/general/#horizontal-form) using bulma which also uses `<label>` for the additional label/title for checkbox/radio groups and doesn't using `<fieldset>`/`<legend>`.

ndm2: > Reading that issue I fail to see why they can't achieve same display while still having valid HTML but let's not get into that.

Styling the `legend` element is really difficult, especially in the context of a `fieldset`, browsers do apply all sorts of annoying restrictions there. The float workaround only works in Firefox, and AFAIK the only somewhat realiable way of positioning would be by using absolute positioning, but that would come with its own issues, especially with regards to responsiveness.

I can't really believe that we still have to deal with this, this behavior is annoying people for decades already, I bet this is in one way or another Internet Explorer's fault, like anything else that sucks on the internet :grimacing: 

Anyways... let's see how using divs and labels works out, I'll give it a try later today.

ADmad:
Nice, I checked with your demo app and the display looks good to me. 

Just for `/pages/default-align#select-checkbox-inline-anchor` the input tags end up with `inline="1"` so the `inline` key needs to be removed/ignored at appropriate location.

ndm2:
Good spot, didn't noticed that, I've completely forgot to add tests for the inline option initially.

If you're OK with the markup, then I'm going to finish up the tests.

ADmad:
@ndm2 Anything more to be done here?

ndm2:
I think we're good for now. Groups in inline forms could need some additional work, but I'd save that for later, as it's really some freaky edge case.

ADmad:
Great, thanks.

PR #245, opened 2018-12-04
Fix duplicate "class" attribute for file inputs.

ADmad:
@ndm2

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -61,7 +61,6 @@ class FormHelper extends Helper
         'inputGroupAddon' => '<div class="{{class}}">{{content}}</div>',
         'inputGroupContainer' => '<div{{attrs}}>{{prepend}}{{content}}{{append}}</div>',
         'inputGroupText' => '<span class="input-group-text">{{content}}</span>',
-        'file' => '<input type="file" class="form-control-file" name="{{name}}"{{attrs}}>',
         'multicheckboxContainer' => '<div class="form-group {{type}}{{required}}" role="group" aria-labelledby="{{groupId}}">{{content}}{{help}}</div>',
         'multicheckboxContainerError' => '<div class="form-group {{type}}{{required}} is-invalid" role="group" aria-labelledby="{{groupId}}">{{content}}{{error}}{{help}}</div>',
         'multicheckboxLabel' => '<label id="{{groupId}}" class="d-block">{{text}}</label>',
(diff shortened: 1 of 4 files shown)


ndm2:
Wouldn't it suffice to use Cakes default file template and inject the `form-control-file` class into the options in `FormHelper::control()`?

ADmad:
> Wouldn't it suffice to use Cakes default file template 

Cake's default file template is being used now.

> and inject the form-control-file class into the options in FormHelper::control()?

That's an option but without overriding the `FileWidget` one won't be able to use the `prepend` and `append` options, hence I injected the class there.

ndm2:
I was wondering about that, doesn't input groups only work properly for these ["custom" file inputs](https://getbootstrap.com/docs/4.1/components/input-group/#custom-file-input)?

ADmad:
Blegh! I hadn't seen that, so yes it does require using `custom-file-input` instead of `form-control-file` and other markup changes for input groups for file input. But I don't want to get into that for now.

Since we'll need even more customizations for input groups we surely need a custom FileWidget class for that.

ndm2:
Hehe, very possible that it will require one :)

The output looks good, only for the horizontal align the label needs an additional `pt-0` class to get it aligned properly. In inline forms the label should probably not be hidden, but AFAICT this will interfere with the block level display applied to the input, so I guess that would also be a job for these custom inputs, ie something for later.

ndm2:
On a second look, maybe it's already aligned properly, or needs a `pt-1`, gotta have a look at it in a few different browsers...

ndm2:
Looks like alignment is optimized for that custom input thingy - yay :)

It's off for regular inputs, which shouldn't be surprising given how differently browsers render that input. In Firefox and IE a `pt-0` looks best, ie Chrome `pt-1` works best.

So maybe that's something users should "fix" on their own...

ADmad:
I have pushed an update to deal with alignment in horizontal form. Chrome/webkit last largest market share so used `pt-1` :slightly_smiling_face:

ADmad:
Looking at the multi checkboxes again I there there too using `pt-1` for the common label/title would be better than `pt-0` you have used.

ndm2:
I think that's kind of an optical illusion, probably because the checkbox position is a little bit off, unlike for the inline variants, in the regular and the nested ones the box doesn't sit exactly on the text baseline.

With `pt-1` however it looks way off for me... maybe there's a browser version/OS difference?

ADmad:
Meh, never mind.

ndm2:
Ok 🙃

ADmad:
You happy with these file input changes?

ndm2:
Yep, looks good to me.

ndm2:
> Blegh! I hadn't seen that, so yes it does require using `custom-file-input` instead of `form-control-file` and other markup changes for input groups for file input. But I don't want to get into that for now.

I'm on to it right now... looks like it's easier than I first thought, I've got all controls mostly working (except for the range thing).

PR #247, opened 2018-12-05
BS4 Compatibility - Checkbox Control

ndm2:
This adds support for nested inputs and IMHO bettter inlining. For horizontal forms, inlining doesn't seem to make sense if each checkbox is going to be in a separate row, hence I've disabled it.

Haven't added any tests yet, I wanted to to get an opinion on this first. The test test app has been updated with additional variants, and it will generate two checkboxes at once to check how they align.


ADmad:
The display looks good.

The `control()` method has got quite big, and it's starting to get difficult to follow the code. At some point we should start splitting it into separate helper methods.

ndm2:
I've already added a little bit more for the date widget, and then there's error and tooltip handling which could possibly further bloat the method... we'll see.

PR #249, opened 2018-12-07
BS4 Compatibility - Datetime Control

ndm2:
This started pretty innocuous, I just wanted to add the `list-inline-item` class to the list elements, so that they are propery aligned.

The widget parts required a separate template so that empty parts do not leave empty list entries, as they are affecting the layout, causing additional space before time only controls, and after date only controls. In order to avoid having to create a separate template for each part, and possibly also its error variants, the part and error class are being injected in the overriden `_datetimeOptions()` method.

The way how an error is being detected might not be overly appreciated, ie checking whether the error class exists in the options. Another way would be checking explicitly for an error via the form context in the `control()` method, and passing the error state into the options, doing that would also allow using an extra error template (`dateWidgetPartError`), and/or using a custom/extended datetime widget, which could also be used for injecting the part class.

No tests yet, but the test app has been updated with additional date widget variants.

On a somewhat related note, that's a lot of additional templates, once we have all controls working, we might want to check if we can reduce them, there's quite some duplication.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -41,10 +41,6 @@ class FormHelper extends Helper
         'dateWidgetPart' => '<li class="list-inline-item {{part}}"><select name="{{name}}"{{attrs}}>{{content}}</select></li>',
         'datetimeContainer' => '<div class="form-group {{type}}{{required}}" role="group" aria-labelledby="{{groupId}}">{{content}}{{help}}</div>',
         'datetimeContainerError' => '<div class="form-group {{type}}{{required}} is-invalid" role="group" aria-labelledby="{{groupId}}">{{content}}{{error}}{{help}}</div>',
-        'dateContainer' => '<div class="form-group {{type}}{{required}}" role="group" aria-labelledby="{{groupId}}">{{content}}{{help}}</div>',
-        'dateContainerError' => '<div class="form-group {{type}}{{required}} is-invalid" role="group" aria-labelledby="{{groupId}}">{{content}}{{error}}{{help}}</div>',
-        'timeContainer' => '<div class="form-group {{type}}{{required}}" role="group" aria-labelledby="{{groupId}}">{{content}}{{help}}</div>',
-        'timeContainerError' => '<div class="form-group {{type}}{{required}} is-invalid" role="group" aria-labelledby="{{groupId}}">{{content}}{{error}}{{help}}</div>',
         'datetimeLabel' => '<label id="{{groupId}}">{{text}}</label>',
         'inputContainer' => '<div class="form-group {{type}}{{required}}">{{content}}{{help}}</div>',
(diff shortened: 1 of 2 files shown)


ADmad:
> The way how an error is being detected might not be overly appreciated, ie checking whether the error class exists in the options. Another way would be checking explicitly for an error via the form context in the control() method, and passing the error state into the options, doing that would also allow using an extra error template (dateWidgetPartError), and/or using a custom/extended datetime widget, which could also be used for injecting the part class.

Yeah checking for a particular class name to decided if the field has errors is pretty flaky. It would be better to use the context methods instead.

> No tests yet, but the test app has been updated with additional date widget variants.

I couldn't get the test app to fetch the updates you have made. After `git pull` I tried both `composer update` and `compose update` but it seems to pull an older commit of BUI. Perhaps the `composer.lock` or the BUI hash in `composer.json` needs to be updated? Anyway, just copied the updated FomHelper from this PR and checked.

The display seems fine. Should there be `:` between the time selects? I don't remember what the core does by default.

> On a somewhat related note, that's a lot of additional templates, once we have all controls working, we might want to check if we can reduce them, there's quite some duplication.

Yeah not too thrilled by all the additional templates required for fixing datetime selects display. I am wondering if it's really worth dealing with this. In practice wouldn't majority of the people using bootstrap just use a JS based datetime picker?  At least that's what I do.

ndm2:
> I couldn't get the test app to fetch the updates you have made.  [...]

Yeah, sorry, I've never updated the composer requirements, you'd need to require the specific commit that you want to test.


> The display seems fine. Should there be `:` between the time selects? I don't remember what the core does by default.

The core doesn't use any separators, but we could sneak them in if you like?

> Yeah not too thrilled by all the additional templates required for fixing datetime selects display. I am wondering if it's really worth dealing with this. In practice wouldn't majority of the people using bootstrap just use a JS based datetime picker? At least that's what I do (through CrudView plugin).

I'm usually using a JS datepicker too, I guess that's true for most people. I don't know, I'd still like to see this working. Given that all the datetime container templates are identical, the `date` and `time` ones could be removed, and we could use the `datetime` ones for all variants, unless the user provides a specific template, hence leaving the ability for the user to use individual templates per variant. That would be 12 templates less.

ndm2:
Ok I just did it and removed the duplicates, please have a look :)

I've tested it with other container templates too, I don't want to speak too soon, because it could easily turn out to be failure, and I've aldready stumbled over one caveat, but it looks like most of them could be removed too.

ADmad:
> Yeah, sorry, I've never updated the composer requirements, you'd need to require the specific commit that you want to test.

No problem, I figured.

> The core doesn't use any separators, but we could sneak them in if you like?

Nah, we don't need to if core doesn't.

> Ok I just did it and removed the duplicates, please have a look :)

Good to see the duplicates are gone. Though there's an issue with display, error message while being generated isn't being shown because it's not matching required CSS selector to set it's display to `block`.

PR #251, opened 2018-12-09
BS4 Compatibility - Datetime Control - Followup

ndm2:
Just when it was merged I figured that I should add some further tests, and that there's a small layout problem that needs to be fixed :)


ndm2:
So what am I going to do with all my new powers now? Save the world?

ADmad:
> Save the world?

By making it perfectly balanced. As all things should be.

ndm2:
Well, I'll start with merging this one myself, but only because it's 99% tests :)

PR #254, opened 2018-12-12
BS4 Compatibility - Custom Form Elements

ndm2:
This should cover all supported custom controls, (multi) checkbox, radio, select, file, and even range... and yes, of course it introduces more templates, more complexity, and more headaches for all the unfortunate people that try to understand it :)

Getting the file label into the input group by rendering it into a template variable in the widget, is the cleanest solution I could come up with so far without making larger structural changes to input group handling in general, which doesn't seem worth the hassle right now, given that currently the file control is the only control that requires this special input group markup.

The controls seem to look and work fine, with a few exceptions (all of which are dictated by bootstrap), nesting inputs isn't supported, checkboxes/radios without labels aren't supported, the file control label is off in inline forms, and the range input doesn't support error/validation feedback.

No tests yet, but the test app has been updated accordingly.

For reference:

* https://getbootstrap.com/docs/4.1/components/forms/#custom-forms
* https://getbootstrap.com/docs/4.1/components/input-group/#custom-forms

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -46,16 +46,24 @@ class FormHelper extends Helper
         'inputContainerError' => '<div class="form-group {{type}}{{required}} is-invalid">{{content}}{{error}}{{help}}</div>',
         'checkboxContainer' => '<div class="form-group form-check {{type}}{{required}}">{{content}}{{help}}</div>',
         'checkboxContainerError' => '<div class="form-group form-check {{type}}{{required}} is-invalid">{{content}}{{error}}{{help}}</div>',
+        'customCheckboxContainer' => '<div class="form-group custom-control custom-checkbox {{type}}{{required}}">{{content}}{{help}}</div>',
+        'customCheckboxContainerError' => '<div class="form-group custom-control custom-checkbox {{type}}{{required}} is-invalid">{{content}}{{error}}{{help}}</div>',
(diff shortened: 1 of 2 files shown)


ADmad:
Clicking on labels for "checkbox (custom, nested)" and "checkbox (custom, inline, nested)" doesn't change the checkbox (visual ?) state.

For horizontal form the label disappears for all "file (custom)" examples.

ndm2:
> Clicking on labels for "checkbox (custom, nested)" and "checkbox (custom, inline, nested)" doesn't change the checkbox (visual ?) state.

That's the expected behavior, bootstrap doesn't support nesting inputs with custom controls. I guess I could quitely disable nesting for custom controls, similar to disabling inlining for horizontal checkboxes.

> For horizontal form the label disappears for all "file (custom)" examples.

Hmm, I can see them. Are you maybe referring to the label that would normally appear in the left column for the regular file control? For the custom ones there is no label on the left hand side, as it's placed inside the control.

ADmad:
> That's the expected behavior, bootstrap doesn't support nesting inputs with custom controls. I guess I could quitely disable nesting for custom controls, similar to disabling inlining for horizontal checkboxes.

Don't think quietly disabling nesting is necessary. Just remove those example from demo app :slightly_smiling_face:.

> Hmm, I can see them. Are you maybe referring to the label that would normally appear in the left column for the regular file control? For the custom ones there is no label on the left hand side, as it's placed inside the control.

Aye, not problem then.

PR #258, opened 2018-12-20
Fix deprecation errors on CakePHP 3.7.

ADmad:


phpstan.neon
```diff
@@ -0,0 +1,4 @@
+parameters:
+    level: 3
```
dereuromark: I liked the -l param more as that one is also visible in the output.
This one is hidden away and more difficult to find, if you just want to know the level.
But no biggie.

src/View/Helper/PaginatorHelper.php
```diff
@@ -14,6 +20,12 @@ class PaginatorHelper extends \Cake\View\Helper\PaginatorHelper
      */
     public function __construct(View $View, array $config = [])
     {
+        if (method_exists($View, 'getRequest')) {
```
dereuromark: These are the downsides of forcing always latest minor via hard deprecations. :/

dereuromark:
Looks good after all.

PR #266, opened 2019-04-06
Do not check for NPM on Windows

ravage84:
This check doesn't work on Windows.

Alternatively one could check by executing npm --version.

### DIFF
--- a/src/Shell/Task/TwbsAssetsTask.php
+++ b/src/Shell/Task/TwbsAssetsTask.php
@@ -38,7 +38,7 @@ class TwbsAssetsTask extends Shell
     public function installAssets()
     {
         $this->info('Checking npm...');
-        if (!`which npm`) {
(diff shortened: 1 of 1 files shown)


src/Shell/Task/TwbsAssetsTask.php
```diff
@@ -38,7 +38,7 @@ public function __construct()
     public function installAssets()
     {
         $this->info('Checking npm...');
-        if (!`which npm`) {
+        if (strtoupper(substr(PHP_OS, 0, 3)) !== 'WIN' && !`which npm`) {
```
ndm2: Maybe use `where.exe npm` on Windows instead?

ravage84: We could also execute ``npm --version`` and check the return value.

PR #272, opened 2019-04-07
Add tool specific files to .gitignore

ravage84:


ADmad:
I wish people would start using global gitignore files for their environment and stop bloating the gitignore files of each project with crap for various IDEs.

ravage84:
@ADmad I get it but for other projects I commit part of the .idea folder, such as the run configurations. Thus I do not want to globally ignore it.

PR #273, opened 2019-04-22
Fixed code to CakePHP 4.x

steinkel:
## Description
Upgrade to CakePHP 4.x


ADmad:
The `develop` branch is for Bootstrap `4.x` and since we haven't had a stable release for `2.0` using this branch yet I am included to just merge these changes in `develop` itself. I personally have no interest in maintaining version of this plugin which support both Cake 3 and Cake 4 with Bootstrap 4.

Anyone with the time and interest could instead update the `master` branch for Cake 4. This would allow people to update their apps to use Cake 4 without being forced to update to Bootstrap 4 too.

src/Shell/BootstrapShell.php
```diff
@@ -66,7 +67,7 @@ public function modifyView($path = null)
      *
      * @return \Cake\Console\ConsoleOptionParser
      */
-    public function getOptionParser()
+    public function getOptionParser(): \Cake\Console\ConsoleOptionParser
```
ADmad: Add "use" statement and just use class name here, not FQCN. Similar change is required in other files too.

composer.json
```diff
@@ -16,12 +16,16 @@
...
-        "cakephp/cakephp": "^3.7.0"
+        "cakephp/cakephp": "4.x-dev as 4.0.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7.14|^6.0",
-        "cakephp/bake": "^1.8"
+        "phpunit/phpunit": "^7",
+        "cakephp/bake": "4.x-dev as 4.0.0",
+        "vlucas/phpdotenv": "^3.3",
+        "cakephp/cakephp-codesniffer": "dev-master"
```
ADmad: This needs to be `dev-next`.

skie: dex-next applied

skie:
ADmad would be nice to get some help from your side on rest CS and phpstan issues.

src/View/Helper/BreadcrumbsHelper.php
```diff
@@ -57,7 +58,7 @@ public function add($title, $url = null, array $options = [])
     /**
      * {@inheritDoc}
      */
-    public function prepend($title, $url = null, array $options = [])
+    public function prepend($title, $url = null, array $options = []): \Cake\View\Helper\BreadcrumbsHelper
```
dereuromark: Are those $this?
Then you must remove all typehints, chainable must never be typehinted here, not even with self/static etc works.

skie: you can't remove. may be only in cakephp core.
just because of this

skie: For stan we should fix "Constant DS not found."

skie:
For cs check, we can rename comparisons templates to other extension.

src/View/Helper/BreadcrumbsHelper.php
dereuromark: I added https://github.com/cakephp/cakephp-codesniffer/pull/243 to solve the issue on higher levels. But yeah, we need to fix the core here.

You can declare the constant in the bootstrap and then add this bootstrap file to phpstan via config. See how this is done e.g. in IdeHelper plugin.

dereuromark: I opened a PR for the core: https://github.com/cakephp/cakephp/pull/13150
It was kind of long overdue :)

dereuromark:
You could also exclude certain dirs for cs check that shouldnt be checked.

phpstan.neon
```diff
@@ -2,3 +2,10 @@ parameters:
     level: 5
     ignoreErrors:
         - '#Constant APP not found#'
+        - '#Constant DS not found#'
```
dereuromark: For me it would be cleaner to load a bootstrap file that contains the constants.

src/View/Helper/BreadcrumbsHelper.php
dereuromark: Core should be fixed now. No typehint for chaining methods.

skie: This fix does not work. I take sometime to take a look for solution and it was not type hint issue for mehtod, but wrong typehint in docblock.

So working for core would be keep retrun self both for method and docblock,
E.g.
```
     * @return $this
     */
    public function prepend($title, $url = null, array $options = []): self
```

dereuromark: Does the code contain $this as return value? Then NO, it is not self, neither in doc block nor code.
$this in docblock and no return type. See core fix and referenced code sniffer rule that can help you.

phpstan.neon
skie: Applied.

src/View/Helper/BreadcrumbsHelper.php
skie: Core fix does not work. Take a look at https://travis-ci.org/FriendsOfCake/bootstrap-ui/jobs/524151404

dereuromark: Did u remove all typehints as you should? Doesnt look like it.

skie: I did it locally. Now i pushed this. Same result in stan check.

dereuromark:
I think your problem is the invalid use of {@inheritDoc}
it marks a partial overwrite. which is a lie in your case. 

So either use it without {} or let it refer only to the title and re-add the param/return types here:

```
@return $this
```

This will please the tool.

src/View/Helper/BreadcrumbsHelper.php
```diff
@@ -46,7 +46,8 @@ public function render(array $attributes = [], array $separator = []): string
     }
 
     /**
-     * {@inheritDoc}
+     * @inheritDoc
```
dereuromark: thats not valid together. see my comment.

skie:
What would be recommendation for deprecated File and Folder replacement?

dereuromark:
IMO those are not deprecated. We shouldnt deprecate sth without having a good and intuitive solution.

skie:
psalm show File and Folder as deprecated.

src/View/Helper/BreadcrumbsHelper.php
dereuromark: Please fix the invalid doc block here as per my review, I mentioned this several times before.

src/View/Helper/FlashHelper.php
```diff
@@ -1,5 +1,5 @@
 <?php
-
+declare(strict_types=1);
```
dereuromark: @ADmad Shouldnt there be a newline between this and and namespace?

ADmad: Enforcing a newline after `declare()` would add a newline above the file level docblocks in core (which have the copyright and license tags).

I don't mind either way. The slevomat sniff could be configured accordingly.

ADmad:
`File` and `Folder` classes are deprecated. We can either ignore the deprecations or use `symfony/filesystem` and `symfony/finder` packages instead.

src/View/Helper/FlashHelper.php
dereuromark: I still hate that those declare those additions cannot be added to open tag on file level.. :(
```
<?php declare(strict_types=1);
```
That resolves the issue as the newline then added below is trivial.

dereuromark: https://github.com/rectorphp/rector/blob/master/src/CodingStyle/AfterRectorCodingStyle.php#L1 etc

ADmad: You are gonna "love" PSR12 https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#example.

I wants an empty line between `<?php` and `declare()` :slightly_smiling_face:.

ADmad:
There seem to be a few people using the `develop` branch. I don't want to cause them unnecessary grief by merging the Cake 4 change to it, hence I have created new branch `cake-4-bs-4` and re-targeted this PR to it.

Can you please fix the merge conflicts so the PR can be merged.

ADmad:
@steinkel Ping

composer.json
```diff
@@ -16,12 +16,17 @@
...
-        "cakephp/cakephp": "^3.7.0"
+        "php": ">=7.2.0",
+        "cakephp/cakephp": "4.x-dev as 4.0.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7.14|^6.0",
-        "cakephp/bake": "^1.8"
+        "phpunit/phpunit": "^7",
+        "cakephp/bake": "4.x-dev as 4.0.0",
+        "vlucas/phpdotenv": "^3.3",
```
ADmad: Why do we needs this package?

composer.json
```diff
@@ -16,12 +16,17 @@
...
+    "minimum-stability": "dev",
     "require": {
-        "cakephp/cakephp": "^3.7.0"
+        "php": ">=7.2.0",
+        "cakephp/cakephp": "4.x-dev as 4.0.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7.14|^6.0",
-        "cakephp/bake": "^1.8"
+        "phpunit/phpunit": "^7",
```
ADmad: Should be phpunit `^8.0` now.

src/View/Helper/FormHelper.php
```diff
@@ -135,6 +168,7 @@ class FormHelper extends Helper
      */
     protected $_widgets = [
         'button' => 'BootstrapUI\View\Widget\ButtonWidget',
+        'datetime' => ['BootstrapUI\View\Widget\DateTimeWidget', 'select'],
```
ADmad: `DateTimeWidget` doesn't use `SelectBoxWidget` anymore, so this should be just `'BootstrapUI\View\Widget\DateTimeWidget'`.

skie: updated

composer.json
skie: removed

composer.json
skie: done

PR #275, opened 2019-05-10
Fixed Button Outline not rendered correctly

### LINKED ISSUE #274: Button Outline not rendered correctly
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* BootstrapUI Version: dev-develop (for Bootstrap 4.0.0)
* Bootstrap Framework Version: 4.0.0
* jQuery: 3.3.1
* Platform and Target: Web, But not really relevant

### What you did
I ran following code
```php
// Form Helper create, some HTML, etc...

<?= $this->Form->button(
    __('Submit'),
    // I want to have an outline  button in primary as block
    ['class' => 'btn btn-outline-primary btn-block']
); ?>

// Some HTML and then form end.
```

### Expected Behavior
I should get a button like this

```HTML
<button class="btn btn-outline-primary btn-block" type="submit">Submit</button>
```

### Actual Behavior
The rendered result is 

```HTML
<!-- btn btn-secondary gets added "magically" -->
<button class="btn btn-outline-primary btn-block btn-secondary" type="submit">Submit</button>
```

### Resolution
I did some debugging and found that the [`OptionsAwareTrait`](https://github.com/FriendsOfCake/bootstrap-ui/blob/develop/src/View/Helper/OptionsAwareTrait.php#L48-L59) adds the classes in Line [43-46](https://github.com/FriendsOfCake/bootstrap-ui/blob/develop/src/View/Helper/OptionsAwareTrait.php#L52) because the `$allClasses` does not contain either `btn`, `btn-block` or `btn-outline-*` .
This is a real bug that shouldn't happen. I think that the `$allClasses` should be extended with the `btn-outline-*` properties.

D4rkMindz:
## Description
This pull request resolves #274 

## Summarize

 - Added additional loop to generate class names for btn-outline  …
   This loop is in the `OptionsAwareTrait::genAllClassNames()` method because it needs to behave always 
 the same. btn-outline-* is also as valid as btn-* therefore it needs to be generated too if the btn-* classes are generated.
- Updated options aware trait test to match the expected output
- Added "enum" types Classes and Element
  Types are better to read and understand (compared to strings). They also decrease (if required) the time you need to refactor code. Its a common best practice to have them.
   - Added tests for types

## Related Issues

fixes #274

### DIFF
--- a/src/View/Helper/Types/Classes.php
+++ b/src/View/Helper/Types/Classes.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace BootstrapUI\View\Helper\Types;
+
+use ReflectionClass;
+use ReflectionException;
+
+/**
+ * Classes
+ */
+class Classes
+{
+    /** @var string The primary class */
+    const PRIMARY = 'primary';
+    /** @var string The secondary clss */
+    const SECONDARY = 'secondary';
+    /** @var string The success class */
+    const SUCCESS = 'success';
+    /** @var string The danger class */
+    const DANGER = 'danger';
+    /** @var string The warning class */
+    const WARNING = 'warning';
+    /** @var string The info class */
+    const INFO = 'info';
+    /** @var string The light class */
+    const LIGHT = 'light';
+    /** @var string The dark class */
+    const DARK = 'dark';
+    /** @var string The link class */
+    const LINK = 'link';
+    /** @var string The small class */
+    const SM = 'sm';
+    /** @var string The large class */
+    const LG = 'lg';
+
+    /**
+     * Get all defined classes
+     *
+     * @return array
+     * @throws ReflectionException On reflection error
+     */
+    public static function getClasses()
+    {
+        $reflection = new ReflectionClass(self::class);
+        $elements = $reflection->getConstants();
+        return array_values($elements);
+    }
+}
--- a/src/View/Helper/Types/Classes.php
+++ b/src/View/Helper/Types/Classes.php
@@ -7,8 +7,10 @@ use ReflectionException;
 
 /**
  * Classes
+ *
+ * All available bootstrap classes
  */
-class Classes
+final class Classes extends Type
 {
     /** @var string The primary class */
     const PRIMARY = 'primary';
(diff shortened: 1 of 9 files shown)


src/View/Helper/Types/Classes.php
```diff
@@ -0,0 +1,49 @@
...
+    /** @var string The large class */
+    const LG = 'lg';
+
+    /**
+     * Get all defined classes
+     *
+     * @return array
+     * @throws ReflectionException On reflection error
+     */
+    public static function getClasses()
```
ADmad: This method is duplicated in both classes. We could instead create a parent class with method `values()` and have `Classes` and `Elements` extend it.

ADmad:
Just FYI you don't need to use "[WIP]" in PR titles anymore. On the pull request form click on the arrow next to "Create pull request" button and it gives you option to "Create draft pull request" :slightly_smiling_face:.

D4rkMindz:
> Just FYI you don't need to use "[WIP]" in PR titles anymore. On the pull request form click on the arrow next to "Create pull request" button and it gives you option to "Create draft pull request" 🙂.

Thanks, i forgot about that feature :)

D4rkMindz:
> Just FYI you don't need to use "[WIP]" in PR titles anymore. On the pull request form click on the arrow next to "Create pull request" button and it gives you option to "Create draft pull request" 🙂.

Thanks, i forgot about that feature :)

src/View/Helper/Types/Classes.php
D4rkMindz: Thank you for bringing this to my attention. I hope it fits now.

src/View/Helper/Types/Type.php
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Type
+ */
+abstract class Type implements TypeInterface
+{
+    /**
+     * Get all constant values of a class
+     *
+     * @return array|mixed
+     * @throws ReflectionException
```
ADmad: Should be `\ReflectionException` else IDEs and static analysers will consider it relative to current namespace.

D4rkMindz: Even though its imported on the top of the file?
This is the way, PHPStorm proposed me to write it.

ADmad:
Thanks

PR #281, opened 2019-08-30
CSS dashboard - lacking arrow 4 asc and desc sort

apricotpoodle:
### Actual Behavior

sorted column headers are not automagically gently displayed
with a small icon (arrow, whatever, …)
Even if in the html code a class "asc" or "desc" is generated,
these class are not considered by BootstrapUI css file.

### Expected Behavior
by default sorted column headers should be displayed with an icon beside.
The classes "asc" and "desc" should be displayed if necessary

## Summarize
I added this code in
vendor/friendsofcake/bootstrap-ui/webroot/css/dashboard.css.

```css
.asc:after {content: " \2193";}
.desc:after {content: " \2191";}
```

## Benefits
This code solves the lack of BootStrapUI css dashboard
and now display gently small icons beside sorted columns header

Related Issues
fixes FriendsOfCake#280

### DIFF
--- a/webroot/css/dashboard.css
+++ b/webroot/css/dashboard.css
@@ -11,6 +11,15 @@ body {
 /*
  * Global add-ons
  */
+/* added default values for sorting classes 
+ * found in style.css display gently small icons 
+ * beside sorted column header 
+ */
+.asc:after {content: " \2193";}
+.desc:after {content: " \2191";}
+/* these ones are funkies */
+/* .asc:after { content: " \e155"; }*/
+/* .desc:after { content: " \e156";}*/
 
 .sub-header {
   padding-bottom: 10px;


dereuromark:
A small screenshot of the output would be nice, for review.

apricotpoodle:
> A small screenshot of the output would be nice, for review.

image
image

dereuromark:
which is what URL/type?

apricotpoodle:
> which is what URL/type?

Not sure to well understand the question.

I didn't change any behavior. 
I just added some class in a css file.
The file dashboard.css.
For displaying the little arrow beside the column title in blue. 
that's all nothing elese.

This the classical behavior, by default that use an GET method url.

So a click on this column title, will toglle the sorting direction

chrome_2019-08-30_16-35-35
 will produce this kind of URL

stere3/something?sort=ELEC_LIB&direction=desc

And respectively

chrome_2019-08-30_16-33-10

will produce 
stere3/something?sort=ELEC_LIB&direction=asc

ADmad:
The arrow isn't properly vertically aligned with the text/link.

dereuromark:
Could be the font type etc.
I think this is better than before already.

@apricotpoodle I just wanted to make sure the direction of arrow fits to desc/asc. But seems so, I am fine with it.

dereuromark:
@apricotpoodle Did you close this to open a new PR?

dereuromark:
If someone wants to further customize, they can still do that on app level, e.g. using fontawesome front icons:
```
th > a.asc:after {
	content: " \f0dd";
	font-family: FontAwesome;
}
th > a.desc:after {
	content: " \f0de";
	font-family: FontAwesome;
}
```

apricotpoodle:
Or if you prefer Glyphicons : 

```css
.asc:after {
    content: " \E155";
    font-family:'Glyphicons Halflings';
}
.desc:after {
    content: " \E156";
    font-family:'Glyphicons Halflings';
}
```

dereuromark:
Feel free to make a follow up PR with screen shots.
If the majority likes it we can merge that one instead (also if the font family is 100% available).

apricotpoodle:
Unfortunately it seems that the old french adage applies here.

> Cela va sans dire, mais mieux vaut le dire !
> It goes without saying, but it's better to say it!
> Ĝi senscie diras! sed estas pli bone diru ĝin !

Indeed, when setting up on my application server the font was only recognized after having specified its name.

So OK,  I'll follow your advice and will propose a new 'PULL REQUEST'
because using Bootstrap without having icons 'fashionable' is not very satisfactory.

Of course these are just aesthetic considerations

PR #303, opened 2020-02-12
Fixes Template issues

challgren:
## Description
Corrects template path to extend during bake and fixed readme path


dereuromark:
Is this ready to go?

challgren:
Yep, I broke down my changes into little bits to make it easier.

PR #305, opened 2020-02-12
Adding docblocks while baking

challgren:
## Description
Adds docblocks when baking templates

## Benefits
Gives users a guide to correctly document their templates

### DIFF
--- a/tests/comparisons/Template/testBakeView.php
+++ b/tests/comparisons/Template/testBakeView.php
@@ -1,3 +1,9 @@
+<?php
+/**
+ * @var \TestApp\View\AppView $this
+ * @var \Cake\Datasource\EntityInterface $article
+ */
+?>
 <?php $this->extend('/layout/TwitterBootstrap/dashboard'); ?>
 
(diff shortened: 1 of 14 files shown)


tests/comparisons/Template/testBakeView.php
```diff
@@ -1,3 +1,9 @@
+<?php
+/**
+ * @var \TestApp\View\AppView $this
+ * @var \Cake\Datasource\EntityInterface $article
```
dereuromark: Does bake use the concrete entity if available?
If not no big deal, I think IdeHelper will auto switch them out once they exist anyway :)

challgren: It uses * @var \{{ entityClass }} ${{ singularVar }} to bake the template so its defined by the bake command

challgren: The tests case doesn't have a entity so I think it's defaulting to \Cake\Datasource\EntityInterface

PR #309, opened 2020-04-11
Fix CS.

dereuromark:
cleanup

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,6 @@
     "authors": [
         {
             "name": "Jad Bitar",
-            "homepage": "http://jadb.io",
             "role": "Author"
         },
         {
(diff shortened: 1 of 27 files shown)


dereuromark:
With this we should make a new minor, raising the min level to 3.7+ and using less deprecations.

composer.json
```diff
@@ -17,10 +16,11 @@
         }
     ],
     "require": {
-        "cakephp/cakephp": "^3.5"
+        "cakephp/cakephp": "^3.7"
```
ADmad, line 19: Why is bumping CakePHP version needed?

ADmad: I have reverted back the Cake version bump. The couple of method exists check in helpers aren't really a problem.

ADmad:
Thanks.

PR #314, opened 2020-05-29
Fix error messages not being displayed with nested input elements

ndm2:
So [**it's been some time**](https://github.com/FriendsOfCake/bootstrap-ui/issues/223#issuecomment-447495234), but there hasn't really been any feedback on how to handle it.

The source of the problem is that the error message element is not a sibling of an (input) element with an `is-invalid` class. This PR fixes that problem in 3 different places, where the input element is nested in a div:

1. In input groups
2. In horizontally aligned checkboxes
3. In custom file controls

For input groups, Bootstrap now [**officially suggests**](https://github.com/twbs/bootstrap/pull/30180) to put the validation response as a sibling to an input group with an `is-invalid` class, so I thought it's safe to go that route, and add the error class to the input group wrapper. Before/After:

input-group-before-after

In horizontally aligned checkboxes, the message (and help text) is moved into the nesting div, which markup wise and visually will unify with default aligned checkboxes. Before/After:

horizontal-checkbox-before-after

For custom file controls I've chosen adding the error class to the wrapper div too, just like for input groups, because a) moving the message element into the wrapper requires some bigger changes with additional templates, and b) it produces a visual glitch as Bootstrap by default restricts the elements height, causing the error message to "overlfow" and being placed over other elements. Before/After:

custom-file-before-after

Problem with message in the wrapper:

custom-file-overflow-problem

A word of warning, none of these fixes are compatible with [**custom client-side validation**](https://getbootstrap.com/docs/4.1/components/forms/#custom-styles), but it didn't work before either, so there's that.

ps. The tests keep growing and growing, and I think I could combine the tooltip and error tests, and also throw in help text usage, that would reduce the clutter a bit, but I'd do that in a followup, I need some rest first, putting these tests together is really tedious.

/cc @ravage84 Not sure, but IIRC you wanted to be notified about this.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -315,8 +315,6 @@ class FormHelper extends Helper
             case 'datetime':
             case 'date':
             case 'time':
-                $options['hasError'] = $this->_getContext()->hasError($fieldName);
-
                 $options['label']['templateVars']['groupId'] =
                 $options['templateVars']['groupId'] =
                     $this->_domId($fieldName . '-group-label');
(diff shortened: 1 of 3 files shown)


ravage84:
@D4rkMindz please have a look and give feedback if you have any issues.

PR #315, opened 2020-05-31
Add support for tooltip error feedback

ndm2:
This PR adds general support for error [**feedback as tooltips**](https://getbootstrap.com/docs/4.5/components/forms/#tooltips), and makes help text in [**inline aligned forms**](https://getbootstrap.com/docs/4.5/components/forms/#inline-forms) screen-reader only. 

I'll clean it up and add/fix tests once there is a consensus on whether this change makes sense, and depending on whether the approach needs to be changed.

Here's how tooltip feedback generally looks (yes, that overlap is expected, and will appear the same with "vanilla" Bootstrap markup):

error-feedback-default-controls-tooltip

## Tooltip feedback as default for inline forms

Currently error message feedback in inline aligned forms isn't unified, it looks kinda messy and can break the layout, hence this PR makes error tooltips the default for inline align. Making help text screen-reader only compliments this, as help text messes the layout up too.

IMHO this makes inline controls pretty much complete, and RC ready.

Here's a before/after for some of the inline controls:

inline-controls-before-after

The width of the tooltip is bound to the width of the parent element, so people would need to style their forms accordingly, but that's just how the default Bootstrap CSS works.

## Non-static positioning required

One problem with feedback tooltips is that they won't be displayed properly unless the parent element is `relative`, `absolute`, `fixed` or `sticky` positioned, which is simply because the tooltips are using `absolute` positioning with a `top` value of `100%`, which can lead to the tooltip being rendered misplaced, or even completely outside of the viewport.

Quote from [**the Bootstrap feedback tooltip docs**](https://getbootstrap.com/docs/4.5/components/forms/#tooltips):

> [...] Be sure to have a parent with `position: relative` on it for tooltip positioning [...]

I've chosen to handle this by introducing a configurable template variable that translates to one of Bootstraps [**`position-*` utility classes**](https://getbootstrap.com/docs/4.5/utilities/position/). The `formGroupPosition` option (there might be a better name for this) defaults to `relative` (if omitted), and can be configured according to one's needs.

There's probably other ways to solve this, for example simply doing nothing, and requiring people to add CSS rules for `.form-group` according to their needs - probably a reasonable requirement, but the downside would of course be that things wouldn't work ouf of the box.

## Doesn't work well with help text

One more problem is the combination with help text, which will change the height of the containing element, causing the feedback tooltip to be placed underneath the help text, which can look odd. There's not really much that could be done about it, people would need to either not use help text (they could use label tooltips instead), or use custom CSS for the feedback tooltips to position them according to their preference. Here's what it looks like with help text vs without:

error-feedback-tooltips-help-text-compare

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -379,8 +379,8 @@ class FormHelper extends Helper
         ];
         $options = $this->_parseOptions($fieldName, $options);
 
-        $formGroupPosition = $options['formGroupPosition'];
-        $errorStyle = $options['errorStyle'];
+        $formGroupPosition = $options['formGroupPosition'] ?: $this->getConfig('formGroupPosition');
+        $errorStyle = $options['errorStyle'] ?: $this->getConfig('errorStyle');
         $custom = $options['custom'];
         $inline = $options['inline'];
         $nestedInput = $options['nestedInput'];
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -15,6 +15,62 @@ class FormHelper extends Helper
     use OptionsAwareTrait;
 
     /**
+     * The default error style.
+     *
+     * @var string
+     */
+    public const ERROR_STYLE_DEFAULT = 'default';
+
+    /**
+     * The none error style.
+     *
+     * @var string
+     */
+    public const ERROR_STYLE_NONE = 'none';
+
+    /**
+     * The tooltip error style.
+     *
+     * @var string
+     */
+    public const ERROR_STYLE_TOOLTIP = 'tooltip';
+
+    /**
+     * Absolute positioning.
+     *
+     * @var string
+     */
+    public const POSITION_ABSOLUTE = 'absolute';
+
+    /**
+     * Fixed positioning.
+     *
+     * @var string
+     */
+    public const POSITION_FIXED = 'fixed';
+
+    /**
+     * Relative positioning.
+     *
+     * @var string
+     */
+    public const POSITION_RELATIVE = 'relative';
+
+    /**
+     * Static positioning.
+     *
+     * @var string
+     */
+    public const POSITION_STATIC = 'static';
+
+    /**
+     * Sticky positioning.
+     *
+     * @var string
+     */
+    public const POSITION_STICKY = 'sticky';
+
+    /**
      * Set on `Form::create()` to tell if the type of alignment used (i.e. horizontal).
      *
      * @var string|null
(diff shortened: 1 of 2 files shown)


ADmad:
Tests are failing.

ndm2:
> Tests are failing.

> I'll clean it up and add/fix tests once there is a consensus on whether this change makes sense, and depending on whether the approach needs to be changed.

😉

ADmad:
> I'll clean it up and add/fix tests once there is a consensus on whether this change makes sense, and depending on whether the approach needs to be changed.

Sorry I overlooked that :slightly_smiling_face:.

Checking https://getbootstrap.com/docs/4.5/components/forms/#tooltips I personally don't like how these error tooltips are displayed, for e.g the tooltip overlaps the label below. So I probably would never use this.

My personal preference aside, supporting this requires additional templates and code and hence  more stuff to grok and maintenance. I don't remember such display being used around the web either so I wonder how many people will actually used this feature and is it worth the additional maintenance.

ndm2:
I was expecting some push back on this, hence I didn't write up the tests yet :)

I can see how it might not be overly appealing, and personally I probably wouldn't use it in regular forms either, but for inline forms I definitely would, IMHO it's a nice way to not mess up the layout and still be able to present additional feedback in the form.

I'm not married to it though, I just can't really think of any other nice way for the plugin to handle error feedback in inline forms out of the box... Of course one could always do it "manually" by overriding the `error` template, and using CSS to handle the necessary `form-group` positioning, it's easy enough if you know what you're doing.

dereuromark:
Maybe less options, a more opininated default might be a simple compromise?

ndm2:
Like dropping the two options, and making relative positioning and tooltips for inline forms the default?

That would reduce some code, but the additional templates would remain, and it might need a few more templates unless the template variable approach for injecting the positioning class is kept.

ADmad:
Meh, I guess there's isn't a reasonable alternative to the changes currently done. Please fix the tests.

ADmad:
The `control()` method has become pretty bloated now, it's should be broken up into smaller helper methods which return the modified `$options` array for each feature. But that's something for a follow up PR.

ndm2:
> Please fix the tests and we can wrap this up.

Will do, but it might take a few days till I can get to it.

ndm2:
Done.

I figured a handful of tests would be enough for the error style feature, as it isn't really control specific, rather it's template specific, and the template usage in the various controls is already covered by the general error tests.

I've also removed the `none` error style, as it doesn't work well with the base form helper, which uses the generated error markup to determine whether to add the error class to the input group wrapper.

Furthermore I've made the `errorStyle` and `formGroupPosition` options overridable helper options, as changing the style globally (respectively per form) would be the most common use case, I can see only very few situations where one would want to use multiple error styles in a single form.

PR #316, opened 2020-06-11
Fixed description

D4rkMindz:
Fixed version of bootstrap in description


dereuromark:
The description is for all branches, best to say 3/4

ADmad:
Better yet just remove both TWBS and CakePHP versions from the descriptions. They can be specified in READMEs of respective branches.

PR #318, opened 2020-06-17
Break up the giant control() method in helper methods.

ADmad:


ADmad:
The diff making the changes look lot more complicated then they are. I just extracted the options manipulation code for each input/feature type into separate methods.

src/View/Helper/FormHelper.php
```diff
@@ -410,183 +414,286 @@ public function control(string $fieldName, array $options = []): string
...
+                break;
 
-                if ($this->_align === 'horizontal') {
-                    $inline = false;
+            case 'range':
+                if ($options['custom']) {
+                    $options['injectFormControl'] = false;
+                    $options = $this->injectClasses('custom-range', $options);
                 }
+                break;
```
ndm2, line 433: Is that an oversight, or did you leave it because it's so short?

ADmad: The latter :)

PR #320, opened 2020-06-19
Convert shells to commands.

ndm2:
I've ported things mostly as they were, from a functionality perspective there's mainly minor differences here and there. More significant changes are:

* a somewhat different entry point
* dropping the duplicated plugin asset removal/linking functionality
* removing the test for the install command in production mode (the supposed difference when installing assets didn't exist (anymore))

Are failure tests something that should be included? Hard to implement them in integration tests, so it would most likely have to be done by testing the command classes directly, and require some refactoring in order to be able to mock certain parts to simulate failures.


ADmad:
> Are failure tests something that should be included?

Up to you if you think the additional effort is worth it. I personally have never used this shell (which is kinda why I neglected to upgrade it).

src/Command/CopyLayoutsCommand.php
```diff
@@ -0,0 +1,69 @@
...
+class CopyLayoutsCommand extends Command
+{
+    /**
+     * @inheritDoc
+     */
+    public function execute(Arguments $args, ConsoleIo $io)
+    {
+        $io->info('Copying sample layouts...');
+
+        $layoutDir = new Folder(Plugin::path('BootstrapUI') . 'templates' . DS . 'layout' . DS . 'examples');
```
ADmad, line 25: Would be nice to avoid the use of deprecated `Folder` and `File` classes but okay if they are.

ndm2: Any idea why `\Cake\Filesystem\Filesystem` is internal?

I could certainly rewrite everything to use vanilla PHP, but in the end it would pretty much just recreate the functionality that CakePHP ships with, which feels kinda redundant.

ndm2:
> > Are failure tests something that should be included?
> 
> Up to you if you think the additional effort is worth it. I personally have never used this shell (which is kind of why I neglected to upgrade it).

Well, I haven't used it either :)

I'll test it with one of the command tests and see if it's too much hassle.

ndm2:
Ok, I've refactored the commands and added tests for failure conditions. When I was halfway through I thought it might not really be worth the hassle, but I was already half way through, so... here it is :)

I think I should also add support for verbosity, all this NPM noise that's being shown by default seems a tad much.

ndm2:
Hmm, Travis is missing from the checks (but it ran (and failed)), haven't seen that before.

ndm2:
Tests are fixed now, but Travis is still missing.

https://travis-ci.org/github/FriendsOfCake/bootstrap-ui

ADmad:
@ndm2 There ya go :slightly_smiling_face:.

ndm2:
Good good... I've seen that quite a few people had this problem, seemingly after the new design/layout has been introduced.

src/Command/CopyLayoutsCommand.php
ADmad: Just so that we won't have to provide BC promise for it's public API :)  It's unlikely to change though and we can use it in BUI.

ndm2: Alright, I'll switch things to use `Filesystem` and vanilla PHP then.

ndm2: done

ADmad:
@ndm2 Is this good to merge?

ndm2:
I'd think so.

PR #324, opened 2020-07-07
Rename `errorStyle` to `feedbackStyle`.

ndm2:
While the helper currently only supports error state feedback, ie valid state feedback isn't yet supported (https://github.com/FriendsOfCake/bootstrap-ui/issues/223#issuecomment-654885462), switching to a more generic terminology now will probably make things easier further down the road.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -15,25 +15,18 @@ class FormHelper extends Helper
     use OptionsAwareTrait;
 
     /**
-     * The default error style.
+     * The default feedback style.
      *
      * @var string
      */
-    public const ERROR_STYLE_DEFAULT = 'default';
+    public const FEEDBACK_STYLE_DEFAULT = 'default';
 
     /**
-     * The none error style.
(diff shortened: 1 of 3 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -15,25 +15,18 @@ class FormHelper extends Helper
...
-    public const ERROR_STYLE_DEFAULT = 'default';
+    public const FEEDBACK_STYLE_DEFAULT = 'default';
 
     /**
-     * The none error style.
+     * The tooltip feedback style.
      *
      * @var string
      */
-    public const ERROR_STYLE_NONE = 'none';
```
ADmad, line 29: I take it `none` isn't needed?

ndm2: Yeah, I [removed that style](https://github.com/FriendsOfCake/bootstrap-ui/pull/315/commits/680b6411b8bba8a92647deaea967afb22c37fc3f) in the original PR, but forgot the constant.

Issue #335, opened 2020-11-24
Add class to inputContainer


davidspeijer:
This is a (multiple allowed):
* [ ] bug
* [x] enhancement
* [ ] feature-discussion (RFC)

* BootstrapUI Version: 3.0.1
* Bootstrap Framework Version: 4.3.1
* jQuery: 3.5.1
* Platform and Target: 

### What you did
```
<?= $this->Form->control('firstname',
	[
		'placeholder' => __('Enter your first name'),
		'inputContainer' => ['class'=>'col-md-6'],
		'label'=>[
			'text'=>__('First name'),
			'class'=>'small mb-1']
	]) ?>
```

### Expected Behavior
Have the col-md-6 class added to the inputContainer
```
<div class="form-group **col-md-6** text">
	<label class="small mb-1" for="firstname">First name</label>
	<input type="text" name="firstname" placeholder="Enter your first name" inputcontainer="test" id="firstname" class="form-control" value="David" maxlength="50">
</div>
```

### Actual Behavior
inputContainer is not having the class added
```
<div class="form-group text">
	<label class="small mb-1" for="firstname">First name</label>
	<input type="text" name="firstname" placeholder="Enter your first name" inputcontainer="test" id="firstname" class="form-control" value="David" maxlength="50">
</div>
```

dereuromark:
A PR with the suggested fix would be very welcome.

davidspeijer:
To be honest; I have no clue how to fix :-).

ndm2:
Not really a bug, or is it? AFAICT there's just no support for it, the core form helper doesn't support custom container attributes via options either.

I'm wondering, @davidspeijer is there a reason that [**horizontal forms**](https://github.com/FriendsOfCake/bootstrap-ui/tree/3.0.1#horizontal-forms) don't do it for you?

Generally being able to set custom container attributes could be relatively useful, it would help to avoid having to supply custom templates for simple changes, but the overly stuffed templates already aren't overly appreciated, and this would require adding at least two additional template variables to all the container templates, one for the class, as the class attribute already exists in the templates, and one for additional attributes.

Example, templates like these:

```html
<div class="form-group {{type}}{{required}}">{{content}}{{help}}</div>
<div class="form-group {{formGroupPosition}}{{type}}{{required}} is-invalid">{{content}}{{error}}{{help}}</div>
```

would need to become something like this:

```html
[...]

PR #336, opened 2020-11-29
Add support for specifying container attributes.

ndm2:
This allows to make simple changes to the container without having to directly resort to using custom templates.

10 lines of code changes, 10000 bazillion lines of test code - you're welcome :) I have no idea why the diff for the tests is so messed up, it's being displayed fine locally, it's just a bunch of new/duplicated methods.

Theoretically a lot of the assertion stuff could be removed if we'd assume that this feature won't affect the inner HTML/templates, and that the other tests already cover that inner HTML/templates, then it would be enough to just assert the outer opening `<div>`... theoretically.

refs #335

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,6 +413,30 @@ This would generate the following HTML:
 </div>
 ```
 
+### Container attributes
+
+Attributes of the outer control container can be changed via the `container` option, cutting the need to use custom
+templates for simple changes. The `class` attribute is a special case, its value will be prepended to the existing
+list of classes instead of replacing it.
+
+```php
+echo $this->Form->control('title', [
+    'container' => [
+        'class' => 'my-title-control',
+        'data-meta' => 'meta information',
+    ],
+]);
+```
+
+This would generate the following HTML:
+
+```html
+<div data-meta="meta information" class="my-title-control form-group text">
+    <label for="title">Title</label>
+    <input type="text" name="title" id="title" class="form-control">
+</div>
+```
+
 ### Appending/Prepending content
 
 Appending/Prepending content to input groups is supported via the `append` and `prepend` options respectively.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -413,6 +413,30 @@ This would generate the following HTML:
...
 
+### Container attributes
+
+Attributes of the outer control container can be changed via the `container` option, cutting the need to use custom
+templates for simple changes. The `class` attribute is a special case, its value will be prepended to the existing
+list of classes instead of replacing it.
+
+```php
+echo $this->Form->control('title', [
+    'container' => [
```
ADmad, line 424: Core provides [templateVars] options for custom template vars, so I would prefer we don't add more custom top level options https://book.cakephp.org/4/en/views/helpers/form.html#adding-additional-template-variables-to-templates

ndm2: I was kinda expecting that, so let me try to further make the case for this feature.

The point is being able to avoid having to use custom templates for as long as possible. Given how complex the templates already are, it would be nice if users could avoid them until they really need to make larger changes. I've caught myself once or twice already, deliberately avoiding customizing the templates, and instead resorting to wrapping the control in additional plain HTML, and while it's usually easy enough, avoiding that too would be nice.

Furthermore, dynamically adding attributes doesn't work with template variables alone, there would need to be additional logic that transforms the individual variables to a well formed string of HTML attributes.

If those arguments can't warm your heart, then I guess it just shouldn't be, and we'll have to continue customizing templates and extending helpers like the wild caveman that we are :neckbeard:

ADmad: I have a raging fire in my heart for bootstrap which wants to incinerate it, but I reluctantly continue using it as free backend themes for other libs are not easily available.

As long as I am not the one maintaining this options / templates soup, I'll just continue bitching about things from time to time and let you hold the torch :slightly_smiling_face:.

ADmad (approved):

README.md
ndm2: This has been the worst trade deal in the history of trade deals, maybe ever!

ndm2:
I very much forgot about this... and about me being able to merge things.

chrisShick:
@ndm2 I don't think this was ever released as a version?

ndm2:
I thought it was, but indeed, there has been no release ever since. Maybe @ADmad can tag a new release later on once a decision has been made about #348

ADmad:
New release done.

PR #344, opened 2021-02-01
Fix alert positioned wrongly in signin view.

### LINKED ISSUE #342: Alert doesn't have linebreak
This is a (multiple allowed):
* [x] bug
* [ ] enhancement

* BootstrapUI Version: 3.0.1
* Bootstrap Framework Version: 4.3.1
* jQuery: 3.5.0
* Platform and Target: - (localhost)

### What you did
Added UIViewTrait to the AppView. Extended the signin layout in a login view, and created a form in it.

### Expected Behavior
Alert should have linebreak

### Actual Behavior
Alert is shown like that:
Screenshot 2021-01-26 at 11 48 11

ndm2:
Flexbox is by default using row direction, causing the alert and the
form to be aligned horizontally.

The view would look like this now:

image

fixes #342

### DIFF
--- a/webroot/css/signin.css
+++ b/webroot/css/signin.css
@@ -7,6 +7,9 @@ body {
     display: -ms-flexbox;
     display: -webkit-box;
     display: flex;
+    -ms-flex-direction: column;
+    -webit-flex-direction: column;
+    flex-direction: column;
     -ms-flex-align: center;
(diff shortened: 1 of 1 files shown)


ADmad:
Since no tests were harmed in the making on this PR we don't need to wait for Travis to finish :slightly_smiling_face:

Issue #341, opened 2021-01-12
Type date is not implementing class to label


davidspeijer:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* BootstrapUI Version: dev-master
* Bootstrap Framework Version: 4.5
* jQuery: 3.5.1
* Platform and Target: 

### What you did
`<?= $this->Form->control('vet_last_checkup', ['type' => 'date', 'label' => ['text' => __('Last Check'), 'class' => 'small mb-1']]) ?>`

### Expected Behavior
To have a label what has a class of `small mb-1`

### Actual Behavior
```
<div class="form-group date" role="group" aria-labelledby="vet-last-checkup-group-label">
    <label id="vet-last-checkup-group-label">Last Check</label>
    <input type="date" name="vet_last_checkup" id="vet-last-checkup" class="form-control" value="">
</div>
```
No class has been applied to the label. Changing `'type' => 'date'` to `'type' => 'text'` is applying the class correctly.
```
<div class="form-group text"
    <label class="small mb-1" for="vet-last-checkup">Last Check</label>
    <input type="text" name="vet_last_checkup" id="vet-last-checkup" class="form-control">
</div>
```

ndm2:
That's a "problem" that's been present for quite a long time, and for now you can only workaround by using custom templates.

It affects various control types, maybe even most, depending on the form align, as for example the default `label` template for the horizontal alignment looks like this:

```html
<label class="col-form-label %s"{{attrs}}>{{text}}{{tooltip}}</label>
```

The label for the datetime control has similar "problems":

```html
<label id="{{groupId}}">{{text}}{{tooltip}}</label>
```

While it could easily take a `class` attribute, the labels for the other alignments cannot:

```html
<span id="{{groupId}}" class="sr-only">{{text}}{{tooltip}}</span>
```

```html
<label id="{{groupId}}" class="col-form-label %s">{{text}}{{tooltip}}</label>
```

It would be nice if we could get this fixed/supported, not sure though if there's an overly clean way to do it without the code either making assumptions about how the templates look (eg default align label templates not having a `class` attribute present, while other aligns do), or without moving the `class` attributes out of the label templates (which would require changing when/how grid classes are injected).

ndm2:
I'm going to try fixing this after completing the form helper test refactoring.

[...]

PR #346, opened 2021-02-05
Refactor form helper test

ndm2:
Initially this was ment to just split out the various control type tests into separate files, but along the this revealed that there were quite a few uncovered constellations, so this is quite a large change, sorry about that.

The tests for label options will fail, I'll solve them separately for #341.


ndm2:
Looks like I've removed more than I should have, -0.58% coverage... unacceptable 🙃

ndm2:
Ok, I'm done for now.

ndm2:
If there are no objections, I'll merge it as is.

PR #348, opened 2021-05-16
Fix label `class` option not working for various controls.

### LINKED ISSUE #341: Type date is not implementing class to label

ndm2:
I completely forgot about this one, it was sitting on my system for quite a while now 😔 

I tried two different variants, this one, which attempts to fix the problem via (existing) template attributes, and another one that used additional template variables. The latter, while not requiring that much additional logic in the options handlers, turned into a steaming hot mess, with lots of backwards incompatibility, and problems like empty `class` attributes, so while it might not seem so, this is actually the more reasonable approach, even though it also has a small BC problem with custom file control label templates (specifically duplicate `class` attributes).

As a side effect, this change also fixes the problem of `for` attributes missing on labels for date/time controls. Related to that the ARIA group handling stuff is no longer needed and has been removed, as date/time controls do not consist of multiple inputs anymore, and therefore the `for` attribute will work fine. Generating the group ID template variable for date/time controls however is being kept, that way this change should be backwards compatible with custom templates.

fixes #341

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -123,7 +123,7 @@ class FormHelper extends Helper
             '<div{{containerAttrs}} ' .
                 'class="{{containerClass}}form-group {{formGroupPosition}}{{type}}{{required}} is-invalid" ' .
                     'role="group" aria-labelledby="{{groupId}}">{{content}}{{error}}{{help}}</div>',
-        'datetimeLabel' => '<label id="{{groupId}}">{{text}}{{tooltip}}</label>',
+        'datetimeLabel' => '<label{{attrs}}>{{text}}{{tooltip}}</label>',
         'inputContainer' =>
             '<div{{containerAttrs}} ' .
                 'class="{{containerClass}}form-group {{type}}{{required}}">{{content}}{{help}}</div>',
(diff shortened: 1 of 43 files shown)


ndm2:
Well, let's hope for the best... It looks good in my [UI test app](https://github.com/ndm2/bootstrap-ui-test-app), and no fundamental changes were needed for the tests, but you never know.

Issue #349, opened 2021-05-30
Bootstrap 5 Support - Tasks and problems


ndm2:
It seems that overall, supporting BS5 will not require too drastic changes, there's a few things broken, but it's absolutely nothing like the BS3 to BS4 transition, most of the generated HTML seems compatible as-is.

Here's a most likely incomplete list of problems/tasks that I've found so far (some of them might also be problems in the current version, which can be backported later on).

General
---

* [x] Upgrade dependencies (drop jQuery, upgrade Popper to 2.x).
* [x] Screenreader-only content is not being hidden (`sr-only` has been renamed to `visually-hidden`).
* [x] Update custom CSS according to browser support changes (prefixes no longer required, etc).
* [x] Update documentation.

Form Helper
---

* [x] Labels are missing bottom margins (they now require a `form-label` class).
* [x] Controls are missing bottom margins (`.form-group` has been removed).
* [x] Inline multi checkbox groups are not inline aligned (`.form-inline` has been removed).
* [x] Remove custom controls, the "custom" style is now the default.
* [x] File input not rendering correctly (`.form-control-file` has been removed).
* [x] Range control labels are vertically misaligned in horizontal forms.
* [x] Select controls should use `form-select` instead of `form-control`.
* [x] Input group addons no longer use a wrapper (`.input-group-append` and `.input-group-prepend` have been removed).
* [x] Tooltips do not toggle anymore (`data-toggle` has been renamed to `data-bs-toggle`).
* [x] Combined inline controls collapse into each other (`.form-inline` has been removed).
* [x] Checkbox help text sticks to the right side of the label (`.form-text` does not set block level display anymore).
* [x] Use rows and columns for [inline forms](https://getbootstrap.com/docs/5.0/forms/layout/#inline-forms).
* [x] ~Custom input classes for radio and checkbox controls prevent `is-invalid` from being added (problem is in core radio widget, it merges options using `+`).~ Not worth the hassle right now.
* [x] Check if and where `pt` classes are still needed... they _are_ needed for range, radio and multi checkbox.
* [x] ~Remove no longer available `form-group` and `form-inline`?~ Keeping them should do no harm, in fact, they make it easier for people to apply their own styles.
* [x] ~Remove seemingly no longer needed `d-block` classes for labels?~ Actually still required.
* [x] Remove backwards compatibility tests?
[...]

PR #350, opened 2021-05-30
Add BS5 support

ndm2:
As hinted in #349, this PR isn't finished yet, depending on the feedback I'd resolve additional outstanding points and add to this PR.

That being, said, what's already included, is ready to be reviewed. It looks like a lot, but the largest part of the changes is in the tests, and that's mostly a lot of small snippets, like added/removed/moved classes and slight markup structure changes.


ndm2:
I've also updated my [UI test app](https://github.com/ndm2/bootstrap-ui-test-app/tree/bui-bs5) for this, however somehow composer isn't able to require commit hashes anymore for me, but that's a problem for tomorrow.

ndm2:
The flexbox attempt for inline forms wasn't the proper way, so if anyone's looking at it, just ignore that for now, I'll try to fix that next [using rows/cols](https://getbootstrap.com/docs/5.0/forms/layout/#inline-forms).

ndm2:
So I've decided to go for [auto-sizing](https://getbootstrap.com/docs/5.0/forms/layout/#auto-sizing) for inline forms, using columns that stack based on window size felt wrong.

However I'm not going to check off that point, as people might have thoughts about this solution.

templates/bake/Template/login.twig
```diff
@@ -6,14 +6,15 @@
 ?>
 <?php $this->extend('/layout/TwitterBootstrap/signin'); ?>
 
-<?= $this->Form->create(${{ singularVar }}, ['class' => 'form-signin text-center']) ?>
+<?= $this->Form->create(${{ singularVar }}, ['class' => 'form-signin']) ?>
 <?= $this->Html->image('BootstrapUI.baked-with-cakephp.svg', ['class' => 'mb-4', 'width' => '250']) ?>
 <h1 class="h3 mb-3 font-weight-normal"><?= __('Please sign in') ?></h1>
-<?= $this->Form->label('email', __('Email address'), ['class' => 'sr-only']) ?>
+<?php if (isset($this->Flash)) echo $this->Flash->render(); ?>
```
ADmad, line 12: Helpers can be auto loaded and many users might not explicitly include them in helpers list, so I don't think the `isset()` check is a good idea.

ndm2: Yeah, I was copypasta-ing it from the default layout, without bothering to check if there actually was a reason why it was written like that.

ndm2:
AFAICT switches are just checkboxes with one additional class, so I've chosen to go with using a template variable instead of adding new templates. If anyone would like to see this solved using additional templates instead, speak now or forever hold your peace!

ndm2:
Here goes color controls. Why the concrete `color()` method you may ask? Because just like some of the other controls it doesn't require extra markup, just an additional class, so this way people can also create them separately, just like with `text()`, `select()`, etc.

They have [problems with the auto-sizing inline forms](https://github.com/twbs/bootstrap/issues/34195), and just like with [the validation problem](https://github.com/twbs/bootstrap/issues/34151) I hope that it's going to be fixed in soon. If there's no progress some time soon, I'll maybe try to fix it myself.

ndm2:
In true _just do it and apologize later_ fashion, here's Bootstrap Icons, FontAwesome as the default is gone, and the Icon font assets are now being included by default, making them work out of the box. Using FontAwesome (or any other icon library for that matter) is of course still possible.

Currently Bootstrap Icons do not ship with sizing classes, so I've added a custom CSS file with classes for now. I've borrowed them from FontAwesome, they seem to work fine, and it keeps things familiar. Maybe the people over at Bootstrap are open to adding sizing classes, I might at least open an issue later on.

ps. I'll ignore stickler's line length complaints for now, there's lots of old content hat exceeds the limit, but isn't checked. I'll clean up things once the upgrade is finished.

ndm2:
Here's what it looks like:

image

And what it can look like when used in for example flash messages:

image

image

ndm2:
And here's icon support for flash messages, so one doesn't have to hack them in with the message, or reimplement the flash element template.

The defaults look like the manually built examples in my previous comment.

ndm2:
Floating labels are neat... they don't work well for select controls in inline forms though, at least in cases where the label text is wider than the widest option text:

image image

There isn't really much that could be done about it, at least not with the current method of laying out inline forms where the controls shrink to the minimum size required, with the labels being absolutely positioned they simply cannot contribute to the controls minimum size.

There's workarounds on the user's end, like specifying a fixed/minimum width for the control container, or building the inline layout manually, but I'm afraid I don't really have any good idea right now how this could be made to work better out of the box.

ndm2:
If anyone's wondering, this is what help text and errors for inlined checkboxes/switches in non-inline layouts looks like:

image
image

It's generally nice to have if one wants it, and allowing this will also make adding support for `aria-describedby` and, possibly custom help classes easier.

dereuromark:
The other night I added autocomplete for bootstrap icons - I opened a PR now ( https://github.com/FriendsOfCake/bootstrap-ui/pull/350 ) for FormatHelper::icon()

I bet the HtmlHelper::icon() in this repo could also benefit from some kind of IdeHelper task that provides optional autocomplete for IDEs like PHPStorm on those magic strings (icon names).
This way you can just get the possible names from typeahead dropdown.

demo

ndm2:
I'm not gonna stop anyone from pushing a PR 😛

ndm2:
Finally! Remember kids, if you just add enough code you'll eventually reach the minimum coverage ratio 😄

image

README.md
```diff
@@ -266,11 +266,28 @@ For those of you who want even more automation, some bake templates have been in
 bin/cake bake [subcommand] -t BootstrapUI
 ```
 
+Currently, bake templates for the following bake subcommands are included:
+
+### `template`
+
+Additionally to the default `index`, `add`, `edit`, and `view` templates, a `login` template is available too. While
+the default CRUD action view templates can be utilised like this:
```
dereuromark, line 274:
```suggestion
the default CRUD action view templates can be utilized like this:
```
AE

README.md
```diff
@@ -930,8 +928,10 @@ Use a different icon set for all flash messages:
 ```php
 $this->loadHelper('Html', [
     'className' => 'BootstrapUI.Html',
-    'iconSet' => 'fas',
-    'prefix' => 'fa',
+    'iconDefaults' => [
+        'iconSet' => 'fas',
```
dereuromark, line 932: should this be then just "set" or "namespace"?
As the prefix isn't needed anymore with the nesting here?

ndm2:
So that's it for now, I'm not going push further changes (except after review feedback of course). If anyone wants to review this big pile of changes in peace - now's your chance :)

README.md
ndm2: I've just documented this wrongly initially, `iconDefaults` was always required. Certainly the option could be renamed, but the `prefix` option is still required - or did you mean something different by "prefix"?

dereuromark: icon prefix i meant. and yes, prefix also doesn't have a prefix :)
So for consistency I wonder if both should be
- set/namespace
- prefix

only

dereuromark: In my icon helper I also allowed prefix to be bool true btw, to auto use the namespace (like fa and fa-icon)
But this is maybe not needed here or out of scope - just wanted to give that input here.

ndm2: Yeah, let's keep the additional functionality for a new PR at this point, I have some more ideas too, like making the set/namespace optional, using a template based solution for easier SVG sprite usage, etc :)

`namespace` sounds good to me.

ndm2:
So if there aren't any objections I'd merge this soon, and then maybe we can release an alpha/preview to possibly garner some feedback.

ndm2:
So here we go... as requested, I'll tag you now that it's merged @ADmad 😛

ADmad:
@ndm2 Do you need to actually do anything? :slightly_smiling_face: You can start the alpha/beta as you see fit.

ndm2:
Not really, I guess... you just said "_tag me when it's merged_" :) If you're cool with it, then I'll draft an alpha pre-release.

dereuromark:
Did you want to try to make a PR for the sandbox, to showcase some live demo cases of it?

ndm2:
Maybe later when things are finalized, I'm not familiar with the Sandbox app yet.

PR #352, opened 2021-06-11
Update jquery version.

### LINKED ISSUE #351: collapse doesn't work with jQuery v3.5.0
TypeError: Cannot convert object to primitive value
while clicking to navbar-toggler button
I fixed it with an upgrade of JQuery to 3.5.1.
https://github.com/twbs/bootstrap/issues/30553
https://blog.jquery.com/2020/05/04/jquery-3-5-1-released-fixing-a-regression/

Thank's

ADmad:
Closes #351

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,9 +8,9 @@
             "integrity": "sha512-rXqOmH1VilAt2DyPzluTi2blhk17bO7ef+zLLPlWvG494pDxcM234pJ8wTc/6R40UWizAIIMgxjvxZg5kmsbag=="
         },
         "jquery": {
-            "version": "3.5.0",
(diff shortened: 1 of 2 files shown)


ndm2:
Not sure how good Bootstrap and Popper are with regards to not breaking things in patch releases, but should we maybe consider dropping the `.lock` file, so that users can easily get the latest patch versions?

ADmad:
Whatever you see fit. I don't use the css/js provideed through the plugin (nor any templates for that matter) so I don't care :slightly_smiling_face:.

ndm2:
Well, I don't want to be the one who breaks everybody's stuff 😄

ADmad:
Everyone lib maintainer does that sooner or later 😂.

ndm2:
Yeah, well... let's leave that for another day 🙃 Maybe we can give it a try with for BS5.

PR #359, opened 2022-02-15
Add support for configurable spacing.

ndm2:
This adds support for configuring the spacing class used by the form helper, as suggested in https://github.com/FriendsOfCake/bootstrap-ui/issues/349#issuecomment-1013967857.

I'm still not overly convinced by it, as styling is so easily done in CSS, but I let others be the judge here. Please voice your opinions on it if you have any.

Only a few tests for now, but it should be enough to illustrate the possible usage of the feature.

/cc @gerhard-lechner

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -119,6 +119,13 @@ class FormHelper extends Helper
     protected $_grid;
 
     /**
+     * Set on `Form::create()` to tell the spacing type.
+     *
+     * @var string|false|null
+     */
+    protected $_spacing;
+
+    /**
      * Default Bootstrap string templates.
      *
      * @var array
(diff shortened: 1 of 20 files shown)


ndm2:
Except for CI, test failures are unrelated.

src/View/Helper/FormHelper.php
```diff
@@ -203,7 +210,7 @@ class FormHelper extends Helper
         'multicheckboxLabel' =>
             '<label{{attrs}}>{{text}}{{tooltip}}</label>',
         'multicheckboxWrapper' =>
-            '<fieldset class="mb-3 form-group">{{content}}</fieldset>',
+            '<fieldset class="%s form-group">{{content}}</fieldset>',
```
ADmad, line 213: Hmm.. mixing placeholder types feels weird.

ADmad: Okay

PR #360, opened 2022-02-23
Escape request params.

ADmad:
Mitigate potential XSS issue.

### DIFF
--- a/templates/layout/default.php
+++ b/templates/layout/default.php
@@ -43,7 +43,10 @@ if (!$this->fetch('tb_footer')) {
 /**
  * Default `body` block.
  */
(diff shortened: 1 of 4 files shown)


ndm2:
Although I'm not sure how it would be possible for it to slip in there, Salines said the problem also occurred on other pages like `login`. Just in case, the example layouts should maybe apply the same escaping to the request params.

ADmad:
I too don't see how the problem could occur for other pages besides error pages but still escaped them for all layouts.

ndm2:
Yeah, better safe than sorry I guess. I've tried to reproduce it with non-error pages, but no luck 🤷‍♂️

There's one left hiding in the dashboard's header

https://github.com/FriendsOfCake/bootstrap-ui/blob/d96507cd606d7943deca72f9478074a2cc1f13a8/templates/layout/examples/dashboard.php#L31

ndm2 (approved):

PR #361, opened 2022-02-23
Update tests.

ADmad:


ADmad:
@ndm2 What do you propose we do about the "prefer-lowest" job? It's practically to have a separate set of expectations for Cake < 4.3 due to the addition of ARIA attributes. I don't want to update the core constraints in composer since the code actually works fine for lower versions of 4.x.

ndm2:
I'd say we inject the two attributes manually if they're not present. I did that for the BS5 version (`aria-describedby` can be omitted):

https://github.com/FriendsOfCake/bootstrap-ui/blob/eb70a72c66dc0e722efd1e7f28a8f69399123711/src/View/Helper/FormHelper.php#L917-L982

PR #366, opened 2022-04-03
Merge BS5 into Master

ndm2:


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.4', '8.0', '8.1']
```
ADmad, line 17: All the changes to ci.yml are regressions.

ndm2: I could've sworn I've chosen "ours" instead of "theirs" 😞

Do we care about a clean history with regards to such conflict resolution mistakes, eg should I fix on top, or should I rather rewrite the history?

ADmad: Do whatever is convenient for you :)

ndm2: Well, first rewrite attempt completely bricked the branch, but second one worked :)

ADmad (approved):

PR #367, opened 2022-04-04
4.x - Automatically generate placeholders for floating labels.

ndm2:
While this will not cover custom types, it should be good enough for most scenarios, eg use of `text` and `textarea` types.

The text inference is copied from the core, but without support for the special `_ids` key, as this would by default cause a `select` input to be generated, where placeholders do not apply.

Refs https://github.com/FriendsOfCake/bootstrap-ui/issues/349#issuecomment-1040063397 /cc @gerhard-lechner

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -5,6 +5,7 @@ namespace BootstrapUI\View\Helper;
 
 use Cake\Core\Configure\Engine\PhpConfig;
 use Cake\Utility\Hash;
+use Cake\Utility\Inflector;
 use Cake\View\Helper\FormHelper as Helper;
 use Cake\View\View;
 use InvalidArgumentException;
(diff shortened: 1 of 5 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -1039,6 +1041,40 @@ protected function _ariaOptions(string $fieldName, array $options): array
         return $options;
     }
 
+    /**
+     * Modify options for placeholders.
+     *
+     * @param string $fieldName Field name.
+     * @param array $options Options. See `$options` argument of `control()` method.
```
dereuromark, line 1048: we usually `array<string, mixed>` those
Same for assoc return value.

ndm2: I'd say that's something for a new PR if we want to adpot that syntax, as there's dozens of places that just use the plain `array` hints.

dereuromark: Well, new code could already do that, the same as you just added return types :)

ndm2: The other option processing methods already have return types, I only copy pasta'd it 😉

ADmad (approved):

PR #371, opened 2022-04-27
Add notice on README for UIViewTrait

umer936:
## Description
Add notice on README for UIViewTrait

## Benefits
Did not realize that the CakePHP FormHelper was being called instead of BootstrapUI's FormHelper as I was adding it to the viewBuilder in my AppController.php

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,11 @@ class AppView extends View
 }
 ```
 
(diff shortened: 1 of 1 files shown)


dereuromark:
Sounds like something we could merge?

ndm2:
The possible clash isn't limited to explicit trait usage though, it would also happen when extending `UIView`.

Issue #370, opened 2022-04-27
Notice of undefined index on using prepend for inputs


umer936:
This is a (multiple allowed):
* [x] bug
* [ ] enhancement
* [ ] feature-discussion (RFC)

* BootstrapUI Version: 4.0.0.
* Bootstrap Framework Version: 5.1.3
* jQuery: N/A
* Platform and Target: N/A (apache on Linux, Firefox web browser)

### What you did
Attempted to use a prepend on an input: 
```
<?php echo $this->Form->input('q1', ['prepend' => 'q1', 'size' => '30', 'value' => $q1]); ?>
```

### Expected Behavior
Echos input field with prepend.

### Actual Behavior
Echos input field with prepend, but has a notice of an undefined lookup.
![image](https://user-images.githubusercontent.com/2078163/165639108-924cfb50-d996-4d84-834e-b923691eb0fe.png)

I will be submitting a PR to fix shortly.

PR #372, opened 2022-04-27
Add default value in InputGroupTrait.php

umer936:
## Description
Adds a default null value for `templateVars` in `InputGroupTrait.php` so 
https://github.com/FriendsOfCake/bootstrap-ui/blob/6924ce44d72af0346998054f715907682210ffd7/src/View/Widget/InputGroupTrait.php#L83 does not print a notice that it is missing. 

Fixes https://github.com/FriendsOfCake/bootstrap-ui/issues/370

### DIFF
--- a/src/View/Widget/InputGroupTrait.php
+++ b/src/View/Widget/InputGroupTrait.php
@@ -35,6 +35,7 @@ trait InputGroupTrait
             'injectFormControl' => true,
             'injectErrorClass' => null,
             'input' => null,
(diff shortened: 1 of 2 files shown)


ndm2:
This should ideally have some small tests for the individual input type methods, similar to eg `FormHelperTest::testAddOnPrependedText()`.

ndm2:
*ping* @umer936 Can you add tests for it? If not, I can add some.

PR #375, opened 2022-06-25
Update CI config

ADmad:


ADmad:
Ugghh.. The addition of `id` attribute to hidden inputs is causing a ton of test failures.

ndm2:
Noice! Maybe we should add scheduled runs so that we get f'd in due time 😃

ADmad:
Also if we add the ids it will fail with ` prefer-lowest`. Personally I wouldn't mind dropping that build.

ndm2:
We could either drop the build, or like with the ARIA changes, implement adding the ID in a backwards compatible manner into the form helper, until we can bump the CakePHP version.

ADmad:
> We could either drop the build, or like with the ARIA changes, implement adding the ID in a backwards compatible manner into the form helper, until we can bump the CakePHP version.

I have just ignored that failure for now.

PR #391, opened 2023-09-27
Fix up null issue

dereuromark:
In my case there was a type error:

> 1) App\Test\TestCase\Controller\MiscControllerTest::testConvertText
TypeError: BootstrapUI\View\Helper\FormHelper::injectClasses(): Argument # 1 ($classes) must be of type array|string, null given, called in /work/devilbox/data/www/sandbox/vendor/friendsofcake/bootstrap-ui/src/View/Helper/FormHelper.php on line 650

Can be avoided by defaulting it to a valid type

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -647,7 +647,7 @@ class FormHelper extends CoreFormHelper
             isset($options['type']) &&
             $options['spacing'] !== false
         ) {
-            $options['container'] = $this->injectClasses($options['spacing'], (array)($options['container'] ?? []));
+            $options['container'] = $this->injectClasses($options['spacing'] ?? [], (array)($options['container'] ?? []));
         }
 
         if (
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -647,7 +647,9 @@ class FormHelper extends CoreFormHelper
             isset($options['type']) &&
             $options['spacing'] !== false
         ) {
-            $options['container'] = $this->injectClasses($options['spacing'] ?? [], (array)($options['container'] ?? []));
+            $classes = $options['spacing'] ?? [];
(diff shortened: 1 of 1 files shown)


dereuromark:
There could also be more wrong from the setup, I will have to check first again

toggenation:
> There could also be more wrong from the setup, I will have to check first again

I just ran into this null bug with a fresh install of CakePHP 5.0.2 and Bootstrap UI 5.0.0

Do we need to re-instate your patch or is there a setup issue that needs tweaking?

ndm2:
@toggenation Do you have a way to reproduce the problem?

The only way I can see this happening is when you're using `FormHelper::control()` without calling `FormHelper::create()` first. This can be fixed, but I'm not sure whether this should generally be expected to work without any side effects.

dereuromark:
Yeah, indeed

echo $this->Form->control('Form.result', ['type' => 'textarea', 'class' => 'halfSize']);

This is breaking existing code

I would be in favor of fixing this, as this is not necessary to blow up this hard IMO

ryanolton:
I've just started migrating a large project from CakePHP 4.5 to 5.0.2 which has many `FormHelper::control()` instances with no corresponding `FormHelper::create()`, and I'm running into this error as well.

I've search the CakePHP documentation and implies that `FormHelper::create()` is required to use `FormHelper::control()`: does anyone know if this is a strict requirement, or merely a suggestion?

The [HTML spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#association-of-controls-and-forms) notes that elements such as `<input>` or `<select>` do not need to be associated with a `<form>`, only that they can be.

ryanolton:
@ndm2 thanks for your help on this, I appreciate it!

dereuromark:
released in https://github.com/FriendsOfCake/bootstrap-ui/releases/tag/5.0.1

PR #403, opened 2024-11-16
implements container.formGroupClass parameter 

skie:
Implements container.formGroupClass parameter to pass or disable classes different from form-group to containers.

When defined complex fields that should have their own containers, now possible to assign the .form-group class at wrapper level, while the inner fields rendered by FriendsOfCake/bootstrap-ui would not have it.

### DIFF
--- a/src/View/Helper/FormHelper.php
+++ b/src/View/Helper/FormHelper.php
@@ -148,27 +148,27 @@ class FormHelper extends CoreFormHelper
             '{{input}}{{label}}',
         'datetimeContainer' =>
             '<div{{containerAttrs}} ' .
-                'class="{{containerClass}}form-group {{type}}{{required}}">{{content}}{{help}}</div>',
+                'class="{{containerClass}}{{type}}{{required}}">{{content}}{{help}}</div>',
         'datetimeContainerError' =>
             '<div{{containerAttrs}} ' .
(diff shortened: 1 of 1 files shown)


src/View/Helper/FormHelper.php
```diff
@@ -657,21 +662,34 @@ protected function _containerOptions(?string $fieldName, array $options): array
...
-        if (!isset($options['container'])) {
-            return $options;
+        if ($skipFormGroupClass) {
+            unset($options['container']['formGroupClass']);
+            if (empty($options['container'])) {
+                unset($options['container']);
+            }
         }
 
         $containerOptions = $options['container'];
```
ADmad, line 674: This line will generate an `undefined index` error if the `unset()` on line 670 is executed. Test case for when `$skipFormGroupClass` ends up being `true` is needed.

P.S. I have fixed the test suite issues on upstream master, so you can rebase.

ADmad:
I am a bit unsure about these changes. This increases the `if/else` conditions and make the code more difficult to follow/maintain in the long run. The idea behind using string templates was avoid this problem.

Instead of appending the `formGroupClass` to `containerClass` how about replacing the `form-group` in the templates with a new template var `formGroupClass`?  Default it's value to `form-group` and override with user provided value if required. I am pretty sure that will require minimal code changes.

skie:
new approach in separate pull requiest https://github.com/FriendsOfCake/bootstrap-ui/pull/406

Issue #418, opened 2025-03-12
App.title


ironbone:
In `dashboard.php` you use
`Configure::read('App.title'),`
but standard instalation of CakePHP 5 does not set `App.title`. So we get an error. Mayby it will be better  to set also a default value there, like:
`Configure::read('App.title', 'BootApp'),`

nxnen8282:
Our system attempted to address this issue:

Added default value 'BootApp' to Configure::read('App.title') calls to prevent errors when App.title is not configured in standard CakePHP 5 installation.

You can examine the suggested modifications here: https://github.com/nxnen8282/FriendsOfCake-bootstrap-ui-418/commit/1dd127bfaef11556a3579f1b16895d85b95d6389.


> [!CAUTION]
> Advisory: This solution was generated using automated tools, and we strongly recommend reviewing the code thoroughly before implementation. The solution may be partial - please consider it as inspiration for your own implementation.

---

http://missionfixinternet.com

PR #423, opened 2025-05-05
Avoid fatal type error

dereuromark:
Resolves https://github.com/FriendsOfCake/bootstrap-ui/issues/418


templates/layout/examples/dashboard.php
```diff
@@ -16,7 +16,7 @@
 <body <?= $this->fetch('tb_body_attrs') ?>>
     <header class="navbar navbar-dark sticky-top bg-dark flex-md-nowrap p-0 shadow">
         <?= $this->Html->link(
-            Configure::read('App.title'),
+            Configure::read('App.title', ''),
```
ADmad, line 19:
```suggestion
            Configure::read('App.title', 'Bootstrap UI'),
```
Without some text the link would be invisible/un-clickable.

Issue #424, opened 2025-06-18
InputGroupTrait : Error with InputGroupTrait when a field (with complex addons options) have a validation error


MaximeOh:
I have an error with a Formless Model when my field have a validation error. 

![Image](https://github.com/user-attachments/assets/60e3777d-abee-4dde-b812-2efc6a000715)
 
Important thing, this issue is present only when my field have addon and addons options with **class** property filled. 

I used CakePhp 5.1.6 and Bootstrap-ui 5.1.1 (the last version)

```php

// no bug here
$this->Form->control('paperWidth', ['class' => 'form-control', 'append' => ['mm']]);

// no bug here too
$this->Form->control('paperWidth', ['class' => 'form-control', 'append' => ['mm', ['size' => 'lg']]]);

// bug here
$this->Form->control('paperWidth', ['class' => 'form-control', 'append' => ['mm', ['size' => 'lg', 'class' => 'input-group-with-unit']]]);

```

MaximeOh:
Yes, I'll try.

PR #425, opened 2025-06-18
Fix the class attribute when the field has add-ons

MaximeOh:
## Description

Fix the class attribute when the field has add-ons (with add-on options) and a validation error occurs.

## Related Issues

#424


ADmad:
Having a test case would be nice.

PR #431, opened 2026-02-19
Fix CI test failures

dereuromark:
## Summary

- Update command help text tests to match CakePHP 5.x output (`-q` flag description changed)
- Fix `InstallCommand` to create `WWW_ROOT` before symlinking (prevents "No such file or directory" error)
- Remove outdated popper.js references from tests (Bootstrap 5.3+ bundles Popper)
- Simplify `testInstallVerbose` to check key messages instead of exact file list output

## CI Status

- **PHP 8.4 & 8.5 highest**: ✅ All tests pass
- **PHP 8.2 highest**: Tests pass, but CI fails due to `--fail-on-all-issues` and a pre-existing `BreadcrumbsHelper` deprecation (not related to this PR)
- **PHP 8.2 lowest**: Shared workflow passes PHPUnit 11/12 options to PHPUnit 10 (workflow issue)
- **cs-stan**: Phive installation issues in shared workflow (workflow issue)

The actual test code is working correctly - the CI failures are due to shared workflow configuration issues that need to be addressed separately.

### DIFF
--- a/src/Command/InstallCommand.php
+++ b/src/Command/InstallCommand.php
@@ -122,12 +122,6 @@ class InstallCommand extends Command
     {
         $io->info('Linking plugin assets...');
 
-        // Ensure the webroot directory exists before symlinking
-        if (!is_dir(WWW_ROOT)) {
-            $filesystem = new Filesystem();
-            $filesystem->mkdir(WWW_ROOT);
-        }
-
         $result = $this->executeCommand(PluginAssetsSymlinkCommand::class, ['name' => 'BootstrapUI'], $io);
         if (
             $result !== static::CODE_SUCCESS &&
--- a/src/Command/InstallCommand.php
+++ b/src/Command/InstallCommand.php
@@ -122,6 +122,12 @@ class InstallCommand extends Command
     {
         $io->info('Linking plugin assets...');
 
+        // Ensure the webroot directory exists before symlinking
+        if (!is_dir(WWW_ROOT)) {
+            $filesystem = new Filesystem();
+            $filesystem->mkdir(WWW_ROOT);
+        }
+
         $result = $this->executeCommand(PluginAssetsSymlinkCommand::class, ['name' => 'BootstrapUI'], $io);
         if (
             $result !== static::CODE_SUCCESS &&
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "cakephp/cakephp": "^5.1.4"
     },
     "require-dev": {
-        "phpunit/phpunit": "^10.5.5 || ^11.5.3 || ^12.4",
+        "phpunit/phpunit": "^10.5.58 || ^11.5.3 || ^12.4",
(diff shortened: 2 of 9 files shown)


dereuromark:
## Remaining CI issues (shared workflow problems)

### 1. PHP 8.2 lowest - PHPUnit version mismatch

**Location:** `ADmad/.github/.github/workflows/testsuite-without-db.yml`

The workflow passes PHPUnit 11/12 CLI options to PHPUnit 10:
```bash
vendor/bin/phpunit --display-phpunit-notices --display-phpunit-deprecations ...
```
These flags don't exist in PHPUnit 10 (`Unknown option "--display-phpunit-notices"`).

**Fix:** Conditionally use different flags based on PHPUnit version, or only use flags common to all supported versions.

---

### 2. PHP 8.2 highest - Deprecation fails CI

**Location:** `ADmad/.github/.github/workflows/testsuite-without-db.yml`

Uses `--fail-on-all-issues` which fails on deprecations. The `BreadcrumbsHelper` has a pre-existing deprecation warning (using `add()` with array instead of `addMany()`).

**Fix options:**
- Fix the `BreadcrumbsHelper` deprecation in bootstrap-ui
- Or adjust the workflow to not fail on framework deprecation warnings

---

### 3. cs-stan - Phive installation fails

**Location:** `cakephp/.github/.github/workflows/cs-stan.yml`

Phive requires TTY confirmation to import GPG keys:
```
Import this key? [y|N]
[ERROR] Needs tty to be able to confirm
```

**Fix:** Use `phive install --trust-gpg-keys <fingerprints>` with the appropriate key fingerprints for phpstan and psalm.

dereuromark:
cc @ADmad - the first two issues are in your shared workflow repo

src/Command/InstallCommand.php
```diff
@@ -122,6 +122,12 @@ public function linkPluginAssets(ConsoleIo $io): void
     {
         $io->info('Linking plugin assets...');
 
+        // Ensure the webroot directory exists before symlinking
+        if (!is_dir(WWW_ROOT)) {
```
ADmad: This seems like something that needs to be mitigated in the tests only. Can't see why webroot directory would not exist in an the actual app.

ADmad:
> the first two issues are in your shared workflow repo

Just need to change phpunit 10's constraint to from `^10.5.5` to  `^10.5.58` to fix that.

dereuromark:
The `^10.5.58` constraint doesn't help here - `--display-phpunit-notices` doesn't exist in PHPUnit 10.x at all, it was introduced in PHPUnit 11.

The shared workflow runs:
```bash
vendor/bin/phpunit --display-phpunit-notices --display-phpunit-deprecations ...
```

These flags don't exist in PHPUnit 10. The workflow would need to detect the PHPUnit version and use different flags, or only use flags that exist in all supported versions.

ADmad:
I have fixed my workflow as well as Cake's to resolve the phive issue.

PR #433, opened 2026-04-12
Update PHP and PHPUnit version requirements

dereuromark:
clarify

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,11 +17,11 @@
         }
     ],
     "require": {
-        "php": ">=8.1",
+        "php": ">=8.2",
         "cakephp/cakephp": "^5.1.4"
     },
     "require-dev": {
-        "phpunit/phpunit": "^10.5.58 || ^11.5.3 || ^12.4 || ^13.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -17,10 +17,11 @@
         }
     ],
     "require": {
+        "php": ">=8.2",
```
ADmad, line 20: The CakePHP constraint is 5.1+ which supports PHP 8.1.  So can't use `>=8.1`?

dereuromark: We are actually already only testing 8.2+
Using 8.1 now makes no sense IMO - we dont even know if it would work still (hidden 8.2 language could be anywhere).

Also those are two different dimensions, you can use Cake 5.1 just fine with 8.2+
All my plugins and apps did too.
See https://github.com/dereuromark/cakephp-ide-helper/blob/083693f2c099663a0bbb047bf76c47a9f82cf839/composer.json#L34-L36

dereuromark: This can be a minor bump of course.
But its no big deal either way, since composer would just not pull it if the setup is not compatible.

ADmad: okay, let's do it with a minor bump

PR #435, opened 2026-05-11
Add Bootstrap 5.3 color mode (data-bs-theme) support

dereuromark:
## Summary

Adds Bootstrap 5.3 color mode (`data-bs-theme`) support — the flagship BS5.3 feature that the plugin currently has zero coverage for (a `grep -rn 'data-bs-theme\|color-mode' src/ templates/ webroot/` against `master` returns nothing).

The new `ColorModeHelper` provides the two pieces an app needs to actually use color modes without writing custom JS:

### `$this->ColorMode->script()` — applies the stored theme on page load

Emits a small inline IIFE that:

- reads the user's choice from `localStorage` (default key: `bs-theme`),
- resolves `auto` against `window.matchMedia('(prefers-color-scheme: dark)')`,
- sets `data-bs-theme` on the configured target (default: `<html>`),
- re-resolves when the OS preference changes while in `auto` mode,
- exposes `window.BootstrapUIColorMode.{get, set}(mode)` for app code,
- dispatches a `bs-theme-changed` DOM event whenever the resolved theme changes.

Placed near the top of `<head>`, it prevents the flash of unstyled theme that every other "add dark mode in the layout" approach has.

### `$this->ColorMode->toggle()` — ready-made switcher widget

Renders a Bootstrap `btn-group` with one button per configured mode (default: light / dark / auto). Clicking a button calls `BootstrapUIColorMode.set()`, toggles `aria-pressed` and the active class, and fires the event. ARIA-labelled, escapes user-supplied labels.

### Usage

```php
// Layout <head>, as early as possible:
<?= $this->ColorMode->script() ?>

// Navbar or wherever the switcher belongs:
<?= $this->ColorMode->toggle() ?>
```

Config knobs (per-call or via `loadHelper`): `storageKey`, `default`, `target`, `modes`, `labels`, `ariaLabel`, `wrapperClass`, `buttonClass`, `activeClass`. Labels go through `__()` if you pass them through; the helper itself escapes them.

### Wiring

- New class: `src/View/Helper/ColorModeHelper.php`
- Auto-loaded by `UIViewTrait` as `$this->ColorMode`.
- `@property` added to `UIView` docblock.
- README has a new section under Usage explaining usage, public JS API, and config keys.

### Notes

- This is a feature, not a fix. Independent of #434 (bug-fix PR) — either can land first.
- No new asset files; the JS is inlined so apps don't have to ship a separate file or symlink anything new.
- No changes to existing helpers — purely additive.
- Existing layouts/examples still ship neutral CSS classes like `bg-light`/`text-dark` in places; those are not touched here. A follow-up could audit `templates/layout/examples/*` and `FlashHelper.php` for hardcoded neutrals if you want full dark-mode parity in the bundled layouts.


src/View/Helper/ColorModeHelper.php
```diff
@@ -0,0 +1,179 @@
...
+ * with.
+ *
+ * Drop `$this->ColorMode->script()` near the top of `<head>` and
+ * `$this->ColorMode->toggle()` wherever you want the switcher (e.g. navbar).
+ *
+ * @extends \Cake\View\Helper<\Cake\View\View>
+ */
+class ColorModeHelper extends Helper
+{
+    public const MODE_LIGHT = 'light';
```
ADmad: Perhaps add an enum instead for the 3 modes.

dereuromark:
Addressed in 2112231. Extracted the three string constants into a `BootstrapUI\View\Helper\ColorMode` backed-string enum. The helper's `default` and `modes` config keys accept either enum cases or plain strings interchangeably so existing apps that pass `'light'`/`'dark'`/`'auto'` keep working, while new code can use the typed API:

```php
use BootstrapUI\View\Helper\ColorMode;

echo $this->ColorMode->toggle([
    'modes' => [ColorMode::Light, ColorMode::Dark],
    'default' => ColorMode::Dark,
]);
```

A small `_modeValue()` helper coerces either form to its string value for rendering. README updated with the enum import. Two new tests cover enum usage on both `modes` and `default`.

src/View/Helper/ColorModeHelper.php
```diff
@@ -0,0 +1,193 @@
...
+        'default' => ColorMode::Auto,
+        // CSS selector for the element that carries `data-bs-theme`.
+        // Default `html` matches the BS5.3 documentation pattern.
+        'target' => 'html',
+        // Modes (and their order) shown in the toggle. Each entry may be a
+        // ColorMode case or a string.
+        'modes' => [ColorMode::Light, ColorMode::Dark, ColorMode::Auto],
+        // Labels shown on the toggle buttons. Keyed by the mode's string
+        // value. Override for i18n.
+        'labels' => [
```
ADmad: Can we make the enum implement `EnumLabelInterface` instead?

src/View/Helper/ColorModeHelper.php
```diff
@@ -0,0 +1,205 @@
...
+        // case or its string value ("light", "dark", "auto").
+        'default' => ColorMode::Auto,
+        // CSS selector for the element that carries `data-bs-theme`.
+        // Default `html` matches the BS5.3 documentation pattern.
+        'target' => 'html',
+        // Modes (and their order) shown in the toggle. Each entry may be a
+        // ColorMode case or a string. ColorMode cases provide their label
+        // via `EnumLabelInterface`; raw strings fall back to `ucfirst($value)`.
+        // Override labels via the standard i18n catalog.
+        'modes' => [ColorMode::Light, ColorMode::Dark, ColorMode::Auto],
```
ADmad: `ColorMode::cases()` gives you this same array, so do we need this option?

dereuromark: What the option actually buys you (verified against the file):
  - Reordering — e.g. [Auto, Light, Dark] to put Auto first.
  - Subsetting — e.g. only [Light, Dark] if an app wants to drop OS-follow.
  - Custom strings — _modeValue / _modeLabel accept ColorMode|string, so apps can register extra
  themes (BS5.3 supports custom data-bs-theme values).

But we can easily default to cases() via null setting here.

dereuromark: Changed the default to `null` and resolve to `ColorMode::cases()` inside `toggle()` so it's no longer a literal duplicate. Kept the config key itself — it's still the seam for apps that want to reorder (e.g. Auto first), subset (drop Auto), or register additional theme strings (BS5.3 supports custom `data-bs-theme` values via CSS). Fixed in 41a7755.

src/View/Helper/ColorModeHelper.php
```diff
@@ -0,0 +1,207 @@
...
+            . ';'
+            . 'var ac=' . json_encode($config['activeClass']) . ';'
+            . 'w.querySelectorAll("[data-bs-theme-value]").forEach(function(b){'
+            . 'b.addEventListener("click",function(){if(window.BootstrapUIColorMode){'
+            . 'BootstrapUIColorMode.set(b.getAttribute("data-bs-theme-value"));}'
+            . 'w.querySelectorAll("[data-bs-theme-value]").forEach(function(x){x.classList.toggle(ac,x===b);});});'
+            . 'if(b.getAttribute("data-bs-theme-value")===active){'
+            . 'b.classList.add(ac);b.setAttribute("aria-pressed","true");}'
+            . 'else{b.setAttribute("aria-pressed","false");}'
+            . '});'
```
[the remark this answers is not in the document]
dereuromark: Fixed in 0f5129b. The click handler now calls a shared `sync(active)` helper that updates both `classList` and `aria-pressed` together (and is invoked once for the initial state too), so assistive tech and visual users see the same state after every click.

src/View/Helper/ColorModeHelper.php
```diff
@@ -0,0 +1,207 @@
...
+     * @param array<string, mixed> $options Per-call overrides for the helper config.
+     * @return string
+     */
+    public function script(array $options = []): string
+    {
+        $config = $options + $this->getConfig();
+
+        $storageKey = json_encode($config['storageKey']);
+        $default = json_encode($this->_modeValue($config['default']));
+        $target = json_encode($config['target']);
```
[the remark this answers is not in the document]
dereuromark: Fixed in 0f5129b. Added `_jsonValue()` which encodes with `JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT | JSON_THROW_ON_ERROR` and routed all inlined values (`storageKey`, `default`, `target`, `activeClass`) through it. Test `testScriptEscapesScriptTagInConfigValues` pins the encoding against a `a</script>b` storageKey.

src/View/Helper/ColorModeHelper.php
```diff
@@ -0,0 +1,207 @@
...
+            . 'var el=document.querySelector(t)||document.documentElement;'
+            . 'var media=window.matchMedia("(prefers-color-scheme: dark)");'
+            . 'function stored(){return localStorage.getItem(k);}'
+            . 'function resolve(m){return m==="auto"?(media.matches?"dark":"light"):m;}'
+            . 'function apply(m){var r=resolve(m);el.setAttribute("data-bs-theme",r);'
+            . 'document.dispatchEvent(new CustomEvent("bs-theme-changed",{detail:{mode:m,resolved:r}}));}'
+            . 'apply(stored()||d);'
+            . 'media.addEventListener("change",function(){if((stored()||d)==="auto"){apply("auto");}});'
+            . 'window.BootstrapUIColorMode={get:function(){return stored()||d;},'
+            . 'set:function(m){localStorage.setItem(k,m);apply(m);}};'
```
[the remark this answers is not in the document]
dereuromark: Half fixed in 0f5129b.

- `localStorage.getItem`/`setItem` are now wrapped in `try/catch` — when storage is blocked the script still applies the default theme rather than throwing and leaving the FOUC it was meant to prevent.
- Skipping the legacy `addListener`/`removeListener` fallback intentionally. The newer `addEventListener` API on `MediaQueryList` is supported by every browser that ships BS5.3 in practice (Safari ≥ 14, shipped 2020-09; Chrome/Firefox/Edge for years). Bootstrap's own color-mode docs use `addEventListener` without a fallback. Adding a legacy branch for browsers older than the framework itself would be dead code.

src/View/Helper/ColorModeHelper.php
```diff
@@ -0,0 +1,207 @@
...
+     * raw string ("light", "dark", "auto", or a custom string for apps that
+     * register additional themes via the `modes` config).
+     *
+     * @param \BootstrapUI\View\Helper\Enum\ColorMode|string $mode Mode value.
+     * @return string Lower-case string identifier (e.g. `'light'`).
+     */
+    protected function _modeValue(ColorMode|string $mode): string
+    {
+        return $mode instanceof ColorMode ? $mode->value : $mode;
+    }
```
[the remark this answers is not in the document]
dereuromark: Fixed in 0f5129b. Dropped the "lower-case" claim from the docblock — raw string inputs are returned as-is so apps registering custom themes (`'High-Contrast'`, etc.) keep their casing. Normalizing here would silently rewrite app-provided values.

dereuromark:
@ADmad and friends? anything missing?
I would like to release a new minor now.

ADmad (approved):

